PR #17, opened 2021-01-05
1-add L block

lukaszspyra:
@pkierat-epam Code with created L-Block ready for review.


src/main/java/com/epam/prejap/tetris/block/LBlock.java
```diff
@@ -0,0 +1,22 @@
...
+            {1, 0},
+            {1, 0},
+            {1, 1},
+    };
+
+    LBlock() {
+        super(IMAGE);
+    }
+
+}
```
pkierat-epam, line 22: No newline at EOF is not recommended.

pkierat-epam:
Points: 20
Nice quality of tests

src/main/java/com/epam/prejap/tetris/block/LBlock.java
lukaszspyra: Did not know, thank you.

PR #18, opened 2021-01-05
Fix bug in generating random moves by RandomPlayer

magzaba:
Last value of Move enum was never reached.
`Method Java.util.Random.nextInt returns a pseudorandom, uniformly distributed int value between 0 (inclusive) and the specified value (**exclusive**), drawn from this random number generator's sequence.`
Before the upper bound was `Move.values().length-1` so the last value from the Move.values() was never generated.


pkierat-epam:
What about tests?

magzaba:
@pkierat-epam Tests added.

src/test/java/com/epam/prejap/tetris/player/RandomPlayerTest.java
```diff
@@ -0,0 +1,28 @@
...
+    public Object[] allMoveEnumValues(){
+        return Move.values();
+    }
+
+    @Test(groups = "Student", dataProvider = "allMoveEnumValues")
+    public void nextMoveShouldGenerateEachPossibleValue(Move move) {
+        Move expected;
+        do {
+            expected= testingRandomizer.nextMove().get();
+        } while(!expected.equals(move));
```
pkierat-epam, line 23: This is a potentially infinite loop. This can be fixed by:
1. Changing RandomPlayer so that Random can be mocked and tested.
2. Using TestNG features instead of the while loop (test invocation count etc.).
3. Changing RandomPlayer so that Random gets a seed and using this to predict next values.
4. Generate enough outputs to test that all options are covered (LEFT, RIGHT, NONE).
Choose the one which fits here the best. Write why.

pkierat-epam:
1. Make sure that everything works as expected w.r.t. changes that were pulled earlier to master. Perhaps you'll need to rebase or merge, or solve conflicts.
1. Currently, master contains another TestNG configuration file. Make sure that yours can be merged with it.

src/test/java/com/epam/prejap/tetris/player/RandomPlayerTest.java
TJB4epam: Just to add - @pkierat-epam speaks for us all. I'll add one more thing - the moment another PR happens, gets reviewed, accepted and merged, then this PR needs to accommodate more changes. :-)

magzaba:
@pkierat-epam   Tests are changed - I've picked 1st approach - mocking Random, because:
1) I don't have to count on probability (how many repetitions are enough to be sure that you get all results)
2) I check exactly those values that I want (test is parameterized, so even when enum is changed, it still covers all values)
3) I could learn how to mock :)

pkierat-epam:
Nicely done mocking. Two final remarks and we can go with it:
* It's a recommended practice to reformat the files (IDEA: `Alt+Shift+L` on Linux or `Ctrl+Alt+L` on Windows; Eclipse: `Ctrl+Shift+F`) that were changed within a given PR. When everyone on the team has the same code formatting settings, it's easier to avoid conflicts caused by white spaces. Please do so. (Don't reformat the whole project though, only the files touched by you.)
* In some files, there are missing new lines at the end. While it doesn't affect the compilation, it might cause any text processing tools (especially various UNIX utils like grep, sed, awk etc.) to fail or generate invalid output, since they expect every line to end with a new line. Below is how to set that in IDEA:
image

magzaba:
@pkierat-epam  Reformatting done.

pkierat-epam:
Points: 14

PR #20, opened 2021-01-05
3-add S block

oskar117:


src/main/java/com/epam/prejap/tetris/block/SBlock.java
```diff
@@ -0,0 +1,19 @@
+package com.epam.prejap.tetris.block;
+
+/**
+ * The class is derived from {@link Block} and represents S-shaped block:
```
pkierat-epam, line 4: There's no need to write what this class extends - it's added automatically by the `javadoc` tool.

src/test/java/com/epam/prejap/tetris/block/BlockFeedTest.java
```diff
@@ -42,4 +43,11 @@ public void shallContainOnlyOneLBlock() {
         //then
         assertEquals(blocks.size(), 1);
     }
+
+    @Test
+    public void blockListContainsSBlock() {
+        BlockFeed blockFeed = new BlockFeed();
+        boolean containsSBlock = blockFeed.blocks().stream().map(Supplier::get).anyMatch(e -> e instanceof SBlock);
```
pkierat-epam, line 50: It would be nice to preserve the layout of the method chain as in the test above. That helps greatly when debugging the code (IDEA shows it quite nicely), You can also comment out a single line in the middle of the chain much more easily.

pkierat-epam: In addition, there was a GWT convention applied in by .your predecessor. As the English proverb says, "When in Rome, do as Romans do".

pkierat-epam: Also, this test asserts a weaker condition than the one above. It only checks the existence of the "S" block, Multiple "S" blocks could affect the random distribution of the blocks - they would be statistically "more frequent". This might make sense in the future as a part of difficulty level raising, but since the previous one was already accepted, it became a requirement, so a similar requirement should be applied here to preserve the balance.

src/test/java/com/epam/prejap/tetris/block/SBlockTests.java
```diff
@@ -0,0 +1,42 @@
+package com.epam.prejap.tetris.block;
+
+import org.testng.Assert;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+@Test(groups = "Block")
+public class SBlockTests {
```
pkierat-epam, line 8: There's already a file with a `*Test` suffix (which is good and common), it's good to follow some naming convention.

TJB4epam:
I've just merged other changes. Make sure you have all you need and that this works after you merge the new master.

PS. When opening a pull-request - DO add a description. Be decent if you hope for a decent review.

oskar117:
I have implemented changes according to your review @pkierat-epam
I apologise for lack of description

src/main/java/com/epam/prejap/tetris/block/BlockFeed.java
```diff
@@ -8,7 +8,7 @@
 
     private final Random rnd = new Random();
     private final List<Supplier<Block>> blocks = List.of(
-        OBlock::new, LBlock::new
+        OBlock::new, SBlock::new, LBlock::new
```
TJB4epam, line 11: Why before the LBlock? Not that I particularly mind, but usually the new items should land at the end?

src/main/java/com/epam/prejap/tetris/block/SBlock.java
```diff
@@ -0,0 +1,19 @@
+package com.epam.prejap.tetris.block;
+
+/**
+ * The class represents S-shaped block:
+ * <pre> ##<br/>## </pre>
```
TJB4epam, line 5: Nice touch. :+1:

src/main/java/com/epam/prejap/tetris/block/SBlock.java
```diff
@@ -0,0 +1,19 @@
...
+
+    private static final byte[][] IMAGE = {
+            {0, 1, 1},
+            {1, 1, 0},
+    };
+
+    public SBlock() {
+        super(IMAGE);
+    }
+}
```
TJB4epam, line 19: I thank you for the tests.

Would you agree that they duplicate the `LBlock` tests? If so - what about CHANGING them in such a way that we have same body of tests (where applicable) and just different data is being fed to those uniform tests?

In the spirit of DRY - Don't Repeat Yourself.

src/test/resources/testng.xml
```diff
@@ -14,6 +14,7 @@
         </groups>
         <classes>
             <class name="com.epam.prejap.tetris.block.LBlockTest"/>
+            <class name="com.epam.prejap.tetris.block.SBlockTest"/>
```
TJB4epam, line 17: Mighty fine but we want groups as well, to be able to run groups / scenarios:

1. check if all is fine with ALL blocks
2. check if all is fine with JUST an S-block
3. just an I-block
4. ...

So please update the grouping of the tests appropriately, as well as the POM.

While we're on the POM side - please update the version - that change brings us up one minor version number.

TJB4epam (changes requested):

TJB4epam (changes requested):
1. POM version still not bumped up.
2. Groups as you define them are an improvement but you're actually defining TESTS rather than GROUPS. Also, I don't have groups for just an IBlock or just an SBlock.

oskar117:
I've applied more suggestions @pkierat-epam or @TJB4epam

src/test/java/com/epam/prejap/tetris/block/BlockFeedTest.java
TJB4epam: @oskar117 care to reply?

src/test/java/com/epam/prejap/tetris/block/SBlockTests.java
TJB4epam: Alek for some reason didn't reply but this is fixed.

src/test/resources/testng.xml
TJB4epam: I see you have made some changes. Would like to hear if this is now done or not? Generally - remarks should have replies. :-)

src/main/java/com/epam/prejap/tetris/block/SBlock.java
oskar117: There are code dublicates already in the repo, so I think that this should be a new issue for that.

src/test/java/com/epam/prejap/tetris/block/BlockFeedTest.java
oskar117: Everything done

src/test/resources/testng.xml
oskar117: I wish to :). Also, should my replies look like "done" if I don't have any questions/comments?

pkierat-epam: "Done" is fine, or you can use emoji reactions if you like. Any form of response is better than silence.

src/test/java/com/epam/prejap/tetris/block/IBlockTest.java
```diff
@@ -5,7 +5,7 @@
 
 import static org.testng.Assert.assertEquals;
 
-@Test(groups = "Block")
+@Test(groups = "IBlock")
```
pkierat-epam, line 8: What is the reason for such change in the group name? This way, we lose the ability to run all the block tests at once. For me, it would be better to leave the original, more general group and just add a specific one. ('groups' is plural after all.)

pkierat-epam (changes requested):
This is the last change that stops me from merging.

src/test/java/com/epam/prejap/tetris/block/IBlockTest.java
oskar117: But there are metagroups defined in testng.xml, so you can run just block group anyway.

oskar117: Oh, I see that it works anyway. Fixed now.

pkierat-epam:
I was just going to merge your PR... when I spotted that after running the app:
```
Time: 00:00:01
+--------------------+
|                    |
|                    |
|          ##        |
|         ##         |
|                    |
|                    |
|                    |
|                    |
|                    |
|                    |
+--------------------+
```
Yes, the O block is slanted... On master, I wasn't able to replicate it. What's really interesting is that all the tests pass :open_mouth:

oskar117:
> I was just going to merge your PR... when I spotted that after running the app:
> 
> ```
> Time: 00:00:01
> +--------------------+
> |                    |
> |                    |
> |          ##        |
> |         ##         |
> |                    |
> |                    |
> |                    |
> |                    |
> |                    |
> |                    |
> +--------------------+
> ```
> 
> Yes, the O block is slanted... On master, I wasn't able to replicate it. What's really interesting is that all the tests pass 😮

Are you sure that it is not just my SBlock? I've run game milion times by now and everything works fine. (Perhaps I can't see something)

TJB4epam:
> >  the O block is slanted... On master, I wasn't able to replicate it. What's really interesting is that all the tests pass open_mouth
> 
> Are you sure that it is not just my SBlock? I've run game milion times by now and everything works fine. (Perhaps I can't see something)

@pkierat-epam  This is just taken from a fetched sources from Aleksander. I lean to @oskar117  explanation. Breaks added by me for readability:

```
Time: 00:00:10
+--------------------+
|         #          |
|        ##          |
|        ###         |

|         ##         |
|        ##          |

|        ##          |
|        ##          |

|         #          |
|         #          |
|         ##         |
+--------------------+
```

We have L, then O, then S - S is the slanted O.

TJB4epam (approved):
Good for merging - @pkierat-epam take a look at my comment w.r.t. the slanted-o block.

PR #26, opened 2021-01-06
19 - fixed .jar file

Szymon-Skazinski:
@pkierat-epam 
pom.xml with fixed mainClass path in maven-jar-plugin


pkierat-epam:
Do you think you could come up with an idea how to write a test for it? (The rule says, the PR won't be merged without a test...)

Szymon-Skazinski:
I assumed (definitely wrong 😉 ) that in this case the tests are unnecessary and it will be difficult to implement them, but from what I can see it is possible, so I will try to implement them ASAP

TJB4epam (changes requested):
Tests desired.

Szymon-Skazinski:
@pkierat-epam 
1) Change pom.xml to change the order of maven plugin execution (package before test)
2) Add JarFileTest - assumed that .jar is correct when the main class name from MANIFEST.FM file equals to Tetris class name

pom.xml
```diff
@@ -33,10 +33,16 @@
...
                             <addClasspath>true</addClasspath>
-                            <mainClass>com.epam.prejap.sdp.git.tetris.Tetris</mainClass>
+                            <mainClass>com.epam.prejap.tetris.Tetris</mainClass>
                         </manifest>
                     </archive>
                 </configuration>
+                <executions>
+                    <execution>
+                        <id>default-jar</id>
+                        <phase>compile</phase>
```
TJB4epam, line 43: It's 11th hour of my workday so perhaps my eyes are playing a trick on me, but are we now making a JAR at a COMPILE stage, where we don't yet have SOURCES compiled?

Jars are built from classes, right?
So, with no classes, how do we put things INSIDE?

Asking for a friend here.

src/test/java/com/epam/prejap/tetris/JarFileTest.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.jar.JarFile;
+import java.util.jar.Manifest;
+
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertNotNull;
+
+@Test(groups = "JarFile")
+public class JarFileTest {
+
+    private static final String jarfilePath = ".\\target\\tetris-0.1.jar";
```
TJB4epam, line 17: Oh poor soul. You sweet summer child.

TJ thinks of ALL the comments he gave here and there about bumping up the versions and all the copious amounts of bug fixes that will flow Dominik's way once the versions WILL be bumped up and this test will fail cause tetris-0.1 will NO longer be just 0.1...

src/test/resources/testng.xml
```diff
@@ -10,12 +10,14 @@
             <run>
                 <include name="Block"/>
                 <include name="Player"/>
+                <include name="JarFile"/>
```
TJB4epam, line 13: Yeah, no.

Unit tests are one, integration tests are OTHER.

No way this is going along with a unit test suite. Please, set this up as an integration test - for now - since Maven lesson about plug-ins is yet to come - feel free to just create a different testng suite file (the XML file) and name it appropriately and put the test there.

src/test/java/com/epam/prejap/tetris/JarFileTest.java
```diff
@@ -0,0 +1,59 @@
...
+
+    @DataProvider
+    public static Object[][] manifestForCheckingJar() {
+        JarFile jarFile = null;
+        Manifest manifest = null;
+
+        try {
+            jarFile = new JarFile(jarfilePath);
+        } catch (IOException e) {
+            System.err.println("Jar file " + e.getMessage() + " doesn't exist");
```
TJB4epam, line 45: Why we don't fail the test here? With a proper fail message? I mean, nobody will even LOOK at error stream since tests are AUTOMATED, right? So, we're after a final result, not what was printed.

Same below, for another try-catch.

src/test/java/com/epam/prejap/tetris/JarFileTest.java
```diff
@@ -0,0 +1,59 @@
+package com.epam.prejap.tetris;
+
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+import java.io.IOException;
+import java.util.jar.Attributes;
+import java.util.jar.JarFile;
```
TJB4epam, line 8: Good thinking with these utils for JAR manipulation. :+1:

src/test/java/com/epam/prejap/tetris/JarFileTest.java
TJB4epam: Hint: build > finalName in POM.

TJB4epam: Additionally, see Path class or use a `/` since IIRC it's corrected by necessary API to whatever's the system uses as a path separator.

TJB4epam (changes requested):

src/test/java/com/epam/prejap/tetris/JarFileTest.java
Szymon-Skazinski: Regarding this topic, i.e. how to deal with throwing an exception in @_DataProvider_, after asking @TJB4epam, he gave me hints on how to deal with it, so I paste it below (in consultation with Tomasz, obviously ;) ) 

1) Have a test that invokes the @_DataProvider_ directly. If the supplier throws an exception, let the test fails, and let other tests depend on him. _(I used this way in my solution)_ 

2) If the @_DataProvider_ throws an exception, let him catch it and convert this exception into specific output data - e.g. a specific manifest with an error message or an empty array and sentence with an exception message, etc. Then you only need to handle it properly in a test based on the data of this provider.

3) Let the @_DataProvider_ return an _Iterator_. Normally the _Iterator_ gives data, but if we have a case that the @DataProvider fails with an exception, we substitute the _Iterator_ whatever it throws.

4) Some other solutions (e.g. from below link)

Links regarding this problem:
https://rolf-engelhard.de/2011/10/fail-instead-of-skip-a-test-when-testngs-dataprovider-throws-an-exception/
https://stackoverflow.com/questions/21284861/testng-is-any-way-where-i-can-mark-test-fail-in-testng-report-and-continue-sam

pom.xml
Szymon-Skazinski: Such a shame 😑  I changed phase to process-classes

src/test/java/com/epam/prejap/tetris/JarFileTest.java
Szymon-Skazinski: Set finalName as ${project.name} ( the sam as artifactId) and use Path class instead of String

src/test/resources/testng.xml
Szymon-Skazinski: New testng suite file added

Szymon-Skazinski:
@TJB4epam 
I have made all the changes related to your review

pom.xml
```diff
@@ -39,10 +40,16 @@
                     <archive>
                         <manifest>
                             <addClasspath>true</addClasspath>
-                            <mainClass>com.epam.prejap.sdp.git.tetris.Tetris</mainClass>
+                            <mainClass>com.epam.prejap.tetris.Tetris</mainClass>
                         </manifest>
                     </archive>
                 </configuration>
+                <executions>
```
pkierat-epam, line 47: There's no need to change Maven's build process to achieve the result in question. Maven has the following phases (incomplete list), executed in order:
1. validate
2. compile
3. test
4. package
5. verify
6. install
7. deploy
Normally, the package is built during the `package` phase. After that, the maven-failsafe-plugin starts integration tests. In order to run your test correctly, you need to:
1. exclude it from the regular `testng.xml` [done]
2. include it in a separate `integration_testng.xml` file [done]
3. configure `maven-surefire-plugin` to use the regular `testng.xml` (see PR from Lukasz B.)
4. configure `maven-failsafe-plugin` to use `integration_testng.xml`
With that, everything should work properly, and this execution block should not be necessary.

pkierat-epam (changes requested):

src/test/java/com/epam/prejap/tetris/JarFileTest.java
pkierat-epam: Why not `${project.artifactId}`?

pom.xml
TJB4epam: See Paweł's remark with regards to Maven execution phases. Once it is resolved, this is also resolved.

Szymon-Skazinski (commented):
@TJB4epam @pkierat-epam 
Update according to review

pom.xml
Szymon-Skazinski: I configured maven-surefire-plugin and maven-failsafe-plugin, but for the latter, I had to configure goals to bind with the corresponding phases, without this, with `mvn verify`, integration tests don't execute

pkierat-epam: :+1:

src/test/java/com/epam/prejap/tetris/JarFileTest.java
```diff
@@ -0,0 +1,52 @@
...
+import java.util.jar.Manifest;
+
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.fail;
+
+@Test(groups = "JarFile")
+public class JarFileTest {
+
+    private static final Path jarfilePath = Path.of(".", "target", "tetris.jar");
```
pkierat-epam, line 18: A non-blocker (just FYI, not to be implmented now):
The problem with such path reference hard-coded in the test is when it changes (for example someone changes `finalName` in the pom.xml file), the test also needs to changed. If you wanted to decouple those two things, you can pass it using [system properties](https://docs.oracle.com/javase/tutorial/essential/environment/sysprop.html). `maven-{surefire,failsafe}-plugin` provide a configuration option called `argLine`, value of which is passed as JVM args to the JVM that runs the tests. Inside `argLine`, you can use maven properties (`${some.property}`).

pkierat-epam (changes requested):
:+1: Bump up the project version in the pom, and you have a GO.

src/test/java/com/epam/prejap/tetris/JarFileTest.java
Szymon-Skazinski: I did it but I'm not sure if it's correct to using pattern `argLine `like that:  `${project.build.directory}/${project.build.finalName}.jar`. What worries me is whether in the case of other operating systems there will be a problem with the path separator, which in this case is hardcoded as `/`.  I tried to use separate `argLine` for `project.build.directory` and `project.build.finalName` but when I try to retrievs those values by System.getProperty it throws NPE

pkierat-epam: On Windows, Java translates '/' to '\' in paths transparently, so no need to worry about that. :+1:

pkierat-epam (approved):

src/test/java/com/epam/prejap/tetris/JarFileTest.java
pkierat-epam: :+1:

pkierat-epam:
All requirements met. Merging.
Points: 12

PR #28, opened 2021-01-07
15 implement random obstacles

BrendanAW:
Extra blocks added based off size of playfield


TJB4epam:
I love the smell of conflicts in the evening, when I'm finding out that... no need to review! ;-P

Or, in other words: please fix the conflicts first.

src/main/java/com/epam/prejap/tetris/game/Playfield.java
```diff
@@ -3,45 +3,94 @@
...
+     * @param isStarterBlocks determines how many blocks Playfield will start with
+     */
+    public Playfield(int rows, int cols, BlockFeed feed, Printer printer, boolean isStarterBlocks) {
         this.rows = rows;
         this.cols = cols;
         this.feed = feed;
         this.printer = printer;
+        this.starterBlockAmt = isStarterBlocks ? (int) (Math.random() * rows / 2) + 1 : 0;
         grid = new byte[this.rows][this.cols];
+        IntStream.rangeClosed(0, starterBlockAmt).forEach(i -> addStarterBlocks());
```
TJB4epam, line 39: Logic in a c-tor is a smell. Here, we have 'n' calls to a method FROM a c-tor. How to test that method then? 
How to test that Playfield was initialized well? All Playfield initialization calls will now also execute 'n' calls to `addStarterBlocks`? Why can't initialization of a starter block set happen outside of this class?

src/main/java/com/epam/prejap/tetris/game/Playfield.java
```diff
@@ -3,45 +3,94 @@
...
+import java.util.stream.IntStream;
+
 public class Playfield {
 
     private final byte[][] grid;
     private final int rows;
     private final int cols;
     private final Printer printer;
     private final BlockFeed feed;
+    public final int starterBlockAmt;
```
TJB4epam, line 17: Public final because?

src/main/java/com/epam/prejap/tetris/game/Playfield.java
```diff
@@ -3,45 +3,94 @@
...
 
-    public Playfield(int rows, int cols, BlockFeed feed, Printer printer) {
+    /**
+     * Sole constructor of Playfield
+     *
+     * @param rows            height of grid
+     * @param cols            width of grid
+     * @param feed            block generator
+     * @param printer         displays grid to user via System.out
+     * @param isStarterBlocks determines how many blocks Playfield will start with
```
TJB4epam, line 30: A boolean is unable to do that due to having 2 possible values which do not match the "how many" :-)

src/main/java/com/epam/prejap/tetris/game/Playfield.java
```diff
@@ -3,45 +3,94 @@
...
+        switch (move) {
+            case LEFT -> moveLeft();
+            case RIGHT -> moveRight();
+        }
+        moved = moveDown();
         show();
         return moved;
     }
 
+    private void addStarterBlocks() {
```
TJB4epam, line 71: I'm still not certain why Playfield. It was a fairly large class already and it had a defined responsibility. Don't you think you're breaking SRP here? Randomizing a set of blocks could well be done outside, with just the result passed in?

I'm looking forward to you testing how large area may be taken by the results of this random placement of random blocks. :sinister-grin: You know, to make sure you're not taking too much space from the player.

src/main/java/com/epam/prejap/tetris/game/Playfield.java
TJB4epam: At least not with `Math.random()*rows/2 +1` involved.

src/test/java/com/epam/prejap/tetris/game/ExtraBlocksTest.java
```diff
@@ -0,0 +1,27 @@
...
+
+import com.epam.prejap.tetris.block.BlockFeed;
+import org.testng.annotations.Test;
+
+import static org.testng.AssertJUnit.assertEquals;
+import static org.testng.AssertJUnit.assertTrue;
+
+@Test
+public class ExtraBlocksTest {
+    int rows = 10;
```
TJB4epam, line 11: Why are all fields package-private?

src/test/java/com/epam/prejap/tetris/game/ExtraBlocksTest.java
```diff
@@ -0,0 +1,27 @@
...
+    int cols = 20;
+    BlockFeed feed = new BlockFeed();
+    Printer printer = new Printer(System.out);
+
+    @Test
+    public void givenNoFlag_gridsEqual_onPlayFieldConstructor() {
+        var p1 = new Playfield(rows, cols, feed, printer, false);
+        assertEquals("PlayField generated without starter blocks, should return 0", 0, p1.starterBlockAmt);
```
TJB4epam, line 19: Were tests the reason for this field being PUBLIC? If so, why not package-private?

TJB4epam: And even better, why not injected via c-tor?

src/test/java/com/epam/prejap/tetris/game/ExtraBlocksTest.java
```diff
@@ -0,0 +1,27 @@
...
+    }
+
+    @Test
+    public void givenFlag_startBlockAmtGTZero_onPlayFieldConstructor() {
+        var p1 = new Playfield(rows, cols, feed, printer, true);
+        assertTrue(p1.starterBlockAmt > 0);
+    }
+}
```
TJB4epam, line 27: You don't plan on having any tests making sure that some block won't take too large a space? No tests that see how this works with different playfields? If I give a mini-playfield, won't this take too much space?

TJB4epam (changes requested):

pkierat-epam:
You made a significant extraction of of the Playfield's logic into a separate component, which is awesome, but... you only provided a test for the Grid class. How do you ensure that the Playfield class works correctly after the refactor?

src/test/java/com/epam/prejap/tetris/game/GridTest.java
```diff
@@ -0,0 +1,87 @@
...
+
+        grid.hide(block);
+        assertTrue("After hiding block, arrays still not equal", Arrays.deepEquals(original, grid.byteGrid));
+    }
+
+    @Test
+    public void givenNewBlock_blockHasMove_afterGridGenerated() {
+        rows = 10;
+        cols = 10;
+        grid = new Grid(feed, rows, cols);
```
pkierat-epam, line 60: From what I've seen so far, you always need to do two things:

1. invoke the constructor,
2. call the `generateGrid` (which changes the sate of the grid).

For me, it's a perfect opportunity to try a factory method (or even a factory class) that returns an object which is ready to use.

pkierat-epam (changes requested):

pkierat-epam:
You know, it's a good practice to at least respond somehow to the remarks posted in the code review... "fixed", "done", or simple :+1: would do.

BrendanAW:
@pkierat-epam, Sorry I wasn't sure if I should do that here or not

src/test/java/com/epam/prejap/tetris/game/GridTest.java
```diff
@@ -0,0 +1,86 @@
...
+    public void givenNewBlock_blockHasMove_afterGridGenerated(Grid grid) {
+        var rows = grid.byteGrid.length;
+        var cols = grid.byteGrid[0].length;
+        var block = feed.nextBlock();
+
+        //potential moves
+        var count = getCount(block, grid);
+
+        //assertion takes calculated area of options and compares to actual options
+        assertEquals(count, ((rows - block.rows() + 1) * (cols - block.cols() + 1)), "Count should be equal to calculated potential moves");
```
pkierat-epam, line 55: > `((rows - block.rows() + 1) * (cols - block.cols() + 1))`

This expression is:
1. repeated few lines below,
2. unclear.

Maybe a method extraction and a human-readable name would help?

src/main/java/com/epam/prejap/tetris/Tetris.java
```diff
@@ -49,9 +50,13 @@ public static void main(String[] args) {
         int delay = 500;
 
         var timer = new Timer(delay);
+
         var feed = new BlockFeed();
         var printer = new Printer(System.out, timer);
-        var playfield = new Playfield(rows, cols, feed, printer);
+        var flagPresent = Arrays.asList(args).contains("-rb");
```
pkierat-epam, line 56: Now, imagine more and more people are adding their options here. (There is at least one person that has something to add.) I'm gonna leave it for now, counting that others will figure out a decent way to work it out.

In case you're interested, there's a nice library called JCommander which helps with parsing command-line options quite nicely.

pkierat-epam (changes requested):

BrendanAW:
@pkierat-epam, The tests have been updated, although I'm uncertain about JCommander, to use it, should I make a zero-arg constructor? If we are to  be parsing the main args this way, would a separate main class not be a better approach? Finally if Martyna's code has already been approved, should I just add a method to check for my flag within her class?

TJB4epam (changes requested):
New PR hast just been merged. Frankly, this is the next item on my list, and I'd have merged it already if you had bumped the POM version and added more Javadocs. I can live without JDs, but bumping the POM is required. Also - refresh your base here, master has changed.

BrendanAW:
@TJB4epam, merging causes an error that I would like to know how to address. The master's changes are configured to check the main args for navigation keys, but if navigation keys are not present and my flag is, it throws an error stopping the program, can I change the class methods to resolve this, or should I simply add logic to work around the checkArgsForNavigationKeys method from being called?

BrendanAW:
@TJB4epam, I fixed the main method the simple way, added a javadoc for Grid and upped the version by 0.1

src/main/java/com/epam/prejap/tetris/game/Grid.java
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    /**
+     * @param feed           supplies grid with new blocks
+     * @param rows           height of byteGrid
+     * @param cols           width of byteGrid
+     * @param addExtraBlocks to determine if blocks will be added
+     * @return new Grid with updatedByteGrid
+     */
+    public static Grid getNewGrid(BlockFeed feed, int rows, int cols, boolean addExtraBlocks) {
```
pkierat-epam, line 39: For future:

Altering the behavior of the method with a boolean flag is usually a symptom of SRP violation (two methods merged into one).

One way to avoid it is to have more than one Grid implementation: a base one (which wouldn't require a feed), and another one extending it with the random blocks... Oh wait, inheritance is prohibited ;-)

So another option is to use the Strategy pattern to pass extra behavior as an argument depending on the '-rb' flag.

BrendanAW:
@pkierat-epam, alright I'm looking into what you mentioned but as of now, my code can be merged, as it works appropriately

pkierat-epam:
Merging.
Points: 125

TJB4epam (approved):

PR #31, opened 2021-01-07
11 implement game speed increase

Tumilok:
Introduced an increase of game speed depending on the score. I've decided to move some logic from the Tetris class to Score and Waiter classes what gave a good opportunity to introduce game speed.


Tumilok:
I have some questions about my implementation, so I would ask them below:
1. I've removed final keyword from both fields of points in the Score class and milliseconds in the Waiter class and I'm not 100% sure about that decision. From the one hand it's much safer with final keywords, but as I wanted to modify the value I had to change that filed. Of course I could go with creating new object every time I'm increasing the score or reducing the time, but I wasn't sure that this is a good solution ether. What is better in this case: remove final keyword or return new object when modifying one of its fields?
2. The second one is about tests. I cannot test some features using just Unit Testing as some of newly created methods are calling other methods from other classes. Should I test it with integration tests or its work for testers? if yes, can you give some tips on how to do it correctly? 
3. The idea of calling the Waiter#decreaseWaitingTime method from the Score class is taken from the observer pattern. In the beginning I wanted to implement the whole observer pattern with an interface and list of its items in the score class, but at some point I've thought that it's overkill for this case and I should not do it. Was it a good idea for that particular project or I should have implemented a proper observer pattern?

src/main/java/com/epam/prejap/tetris/Score.java
```diff
@@ -1,15 +1,41 @@
...
+ */
+
 class Score {
 
-    private final int points;
+    private final Waiter waiter;
+    private int points = 0;
+
+    public Score(Waiter waiter) {
+        this.waiter = Objects.requireNonNull(waiter, "waiter must not be null");
```
TJB4epam, line 20: Nice.

src/main/java/com/epam/prejap/tetris/Score.java
```diff
@@ -1,15 +1,41 @@
...
+    private boolean shouldDecreaseWaitingTime() {
+        return points % 10 == 0;
     }
 
     int points() {
         return points;
     }
-
-}
+}
```
TJB4epam, line 41: Sigh. And we even made an announcement on MSTeams about that... ;-P

src/main/java/com/epam/prejap/tetris/Score.java
```diff
@@ -1,15 +1,41 @@
...
 
-    private final int points;
+    private final Waiter waiter;
+    private int points = 0;
+
+    public Score(Waiter waiter) {
+        this.waiter = Objects.requireNonNull(waiter, "waiter must not be null");
+    }
+
+    public void increaseScore() {
```
TJB4epam, line 23: increaseScore AND handlesWaitingTime... how to call that method, I wonder...

src/main/java/com/epam/prejap/tetris/Tetris.java
```diff
@@ -58,4 +58,4 @@ public static void main(String[] args) {
         System.out.println("Score: " + score.points());
     }
 
-}
+}
```
TJB4epam, line 61: We want a newline at EOF, so please make it so in all cases.

src/main/java/com/epam/prejap/tetris/game/Waiter.java
```diff
@@ -2,9 +2,16 @@
 
 import java.util.concurrent.TimeUnit;
 
+/**
+ * The Waiter class is responsible for waiting some given time
```
TJB4epam, line 6: Please write JDs without the opening "the x class" or "the class" or "the method". When reading a JD one always knows what they are for and superfluous text should be cut. Here, a good word to start with is "Responsible". 

Then, you're able to have one line as a JD and that's nice and succinct.

src/test/java/com/epam/prejap/tetris/ScoreTest.java
```diff
@@ -0,0 +1,35 @@
+package com.epam.prejap.tetris;
+
+import com.epam.prejap.tetris.game.Waiter;
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+@Test(groups = "Tetris")
```
TJB4epam, line 7: Better name for that group please?

src/test/java/com/epam/prejap/tetris/ScoreTest.java
```diff
@@ -0,0 +1,35 @@
...
+
+    @Test
+    public void shallIncreaseScore() {
+        //given
+        Waiter waiter = new Waiter(0);
+        Score score = new Score(waiter);
+
+        //when
+        score.increaseScore();
+        int expectedPoints = score.points();
```
TJB4epam, line 18: Data given landed in when in this test - why? Both this and next line are test arrangements after all?

src/test/java/com/epam/prejap/tetris/ScoreTest.java
```diff
@@ -0,0 +1,35 @@
...
+        int actualPoints = 1;
+
+        //then
+        Assert.assertEquals(actualPoints, expectedPoints);
```
TJB4epam, line 22: Static import please. No need to read assert assert assert all the time in every test. Apply in other places as well.

TJB4epam: Is it just me or is expected points something else? Expected is your DESIRED outcome mate, actual is what really happend. So, actualPoints being set to 1 makes no sense, neither does setting expectedPoints to result of method from the score.

src/test/java/com/epam/prejap/tetris/ScoreTest.java
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    @Test
+    public void shallThrowNullPointerException() {
```
TJB4epam, line 26: I'd like for the name of the test to say more. :-) Like, WHEN or WHY, not just WHAT shall happen. 

For a 1-liner test it's OK to ignore GWT convention. We permit it. ;-)

src/main/java/com/epam/prejap/tetris/Score.java
```diff
@@ -1,15 +1,41 @@
...
+        points++;
+        handleWaitingTime();
+    }
 
-    public Score(int points) {
-        this.points = points;
+    private void handleWaitingTime() {
+        if (shouldDecreaseWaitingTime()) {
+            waiter.decreaseWaitingTime();
```
TJB4epam, line 30: So, this is a good moment to answer one of your questions:


>    The idea of calling the Waiter#decreaseWaitingTime method from the Score class is taken from the observer pattern. In the beginning I wanted to implement the whole observer pattern with an interface and list of its items in the score class, but at some point I've thought that it's overkill for this case and I should not do it. Was it a good idea for that particular project or I should have implemented a proper observer pattern?

Yes. Actual observer patter would've been quite fine here as well. After all, the REASON for decreasing the wait time is DIRECTLY tied to the score as per requirements.

src/main/java/com/epam/prejap/tetris/game/Waiter.java
```diff
@@ -2,9 +2,16 @@
...
+ * till the next render.
+ *
+ * @implNote It gets {@link #milliseconds} filed as a constructor argument
+ */
+
 public class Waiter {
 
-    private final int milliseconds;
+    private int milliseconds;
```
TJB4epam, line 14: Another question you asked us:

>    I've removed final keyword from both fields of points in the Score class and milliseconds in the Waiter class and I'm not 100% sure about that decision. From the one hand it's much safer with final keywords, but as I wanted to modify the value I had to change that filed. Of course I could go with creating new object every time I'm increasing the score or reducing the time, but I wasn't sure that this is a good solution ether. What is better in this case: remove final keyword or return new object when modifying one of its fields?

This was covered during the lectures with immutable classes and is generally considered a very important programming technique: to craft objects so, that they are immutable and a change creates a new copy, rather than changes the old object.

I talked with Paweł - your code may be left as it is (without final, not using the technique I just mentioned above).

src/test/java/com/epam/prejap/tetris/ScoreTest.java
```diff
@@ -0,0 +1,35 @@
+package com.epam.prejap.tetris;
+
```
TJB4epam, line 2: And I'll use this to mention your third question:

>     The second one is about tests. I cannot test some features using just Unit Testing as some of newly created methods are calling other methods from other classes. Should I test it with integration tests or its work for testers? if yes, can you give some tips on how to do it correctly?

Integration tests, Mockito and interaction tests (verifying method A called method B) are options as well. Do remember: integration / interaction tests are on a brittler side. Sometimes one can avoid them by testing the logic... ON A CLASS THAT HAS IT. 

So, if Score calls Waiter telling him to "decrease waiting time" - if you test Score on that logic you need to mock Waiter. But if you test the Waiter on it, you're doing a unit test.

TJB4epam (changes requested):
Very good questions! Answers done while reviewing, so on particular code snippets that were tied to exact nature of the question.

src/main/java/com/epam/prejap/tetris/Score.java
Tumilok: Just to make sure, are you pointing on the broken SRP?

Tumilok: I now that if I name this method - increaseScoreAndHandleWaitingTime I would break the SRP, so what if I just name this method handleScore and it will evaluate increaseScore and handleWaitingTime private methods?

Tumilok:
@TJB4epam I've applied some changes based on your very helpful comments. Should I click something in here to note that I fulfilled all the requested changes or you can approve them without me doing anything?

TJB4epam:
> @TJB4epam I've applied some changes based on your very helpful comments. Should I click something in here to note that I fulfilled all the requested changes or you can approve them without me doing anything?

I'd say "re-request review" is your option (right bottom corner):

obraz

src/main/java/com/epam/prejap/tetris/Score.java
```diff
@@ -1,14 +1,45 @@
...
+ * when player gets next 10 points. It gets {@link Waiter} object as a constructor argument
+ */
+
 class Score {
 
-    private final int points;
+    private final Waiter waiter;
+    private int points = 0;
+
+    public Score(Waiter waiter) {
```
TJB4epam, line 19: Encapsulation asks if she's ditched on purpose and is your new toy better :P

Why a public c-tor on a package-private class?
Why was the method points() made public?
What will be the primary client of a Score class? Where will the calls asking for it's services be?
Is the class located in the right package if it collaborates closely with a class Waiter?

Mind you - there are classes that are smack in the middle of two packages. Is this one such a class?

What if you had a DifficultyLevel, and you would simply raise it? How would this change the whole communication between game, score and waiter?

TJB4epam (changes requested):

Tumilok:
@TJB4epam After thinking for a while about what you said, I decided to implement an observer pattern to reduce dependence between Score and Waiter classes. Now it would not be a problem to introduce new logic connected with the Score class, for instance, Difficulty Level. Also, I was thinking a lot about where to place Score class and how to encapsulate it properly, so I've decided to place it and ScoreObserver interface in its own package named score. I had to make this class package public in order to make possible communication between other packages. I also was thinking about not changing the place of Score class and just making all its fields package-private and move ScoreObserver interface to the place where other classes are implementing it - game package, but in the end, I've decided to go the way I did.

I would be happy if you send feedback about the encapsulation strategy I've chosen. :)

TJB4epam:
Will close this today if conflicts aren't resolved, thank you.

Tumilok:
@TJB4epam I have resolved conflicts

src/test/resources/testng.xml
```diff
@@ -36,8 +41,8 @@
             <class name="com.epam.prejap.tetris.block.IBlockTest"/>
             <class name="com.epam.prejap.tetris.block.BlockFeedTestFactory"/>
             <class name="com.epam.prejap.tetris.player.RandomPlayerTest"/>
-            <class name="com.epam.prejap.tetris.game.MoveTest"/>
-            <class name="com.epam.prejap.tetris.CommandLineAnalystTest"/>
```
pkierat-epam, line 40: What happened to those two tests (MoveTest and CommandLineAnalystTest)? Have they been completely removed? Or maybe there's no point in running them?

src/main/java/com/epam/prejap/tetris/game/Waiter.java
```diff
@@ -16,4 +27,29 @@ public void waitForIt() {
         } catch (InterruptedException ignore) {
         }
     }
+
+    @Override
+    public void scoreChanged(int newScore) {
```
pkierat-epam, line 32: This makes the Waiter dependent on the actual score, while it shouldn't care about the actual value, otherwise you need to change this class as well if the score calculating algorithm changes. A concept of "next level" might be of use here.

pkierat-epam (commented):
My last remarks.

src/main/java/com/epam/prejap/tetris/game/Waiter.java
Tumilok: Just to make sure, "next level" concept is not some kind of pattern or technique, yes? Cause I've already googled it and found nothing :) As I understand I need to move logic of checking score from Waiter class to Score class, am I right?

Tumilok:
@pkierat-epam Reimplement your remarks

pom.xml
```diff
@@ -18,7 +18,7 @@
         <version.plugin.maven.failsafe>3.0.0-M5</version.plugin.maven.failsafe>
         <!-- dependencies versions -->
         <version.testng>7.3.0</version.testng>
-        <version.mockito>3.2.4</version.mockito>
+        <version.mockito>1.10.19</version.mockito>
```
pkierat-epam, line 21: What a downgrade! Is it intentional? :raised_eyebrow:

pom.xml
```diff
@@ -30,7 +30,7 @@
         </dependency>
         <dependency>
             <groupId>org.mockito</groupId>
-            <artifactId>mockito-core</artifactId>
+            <artifactId>mockito-all</artifactId>
```
pkierat-epam, line 33: Related to the version downgrade.

pkierat-epam (changes requested):
Please restore the Mockito version (if not intentional).

pom.xml
Tumilok: @pkierat-epam I'm so sorry! Still learning how to merge files correctly

Tumilok:
@pkierat-epam Fixed bad merging and prepared my part to merge with Paulina's code

Tumilok:
@pkierat-epam When I got merge score wasn't changing during the game. I found a bug in the constructor of Playfiled class, instead of using list of observers from the constructor it just created new empty list. Now everything seems to work just fine. Merge was completed successfully!

pkierat-epam:
Merging.
Points: 18

PR #33, opened 2021-01-08
7 Rotation of blocks

krawmaciej:
@pkierat-epam @TJB4epam 
Hi! I've done the issue #7. I would like your feedback on what should be changed, cheers!


src/main/java/com/epam/prejap/tetris/block/BlockRotator.java
```diff
@@ -0,0 +1,127 @@
...
+
+public class BlockRotator {
+
+    private static final Point NO_OFFSET = new Point(0, 0);
+
+    private Block block;
+    private Point blockOffset;
+
+    private final LastBlockRotation lastRotation;
+    private boolean rotationBackLocked;
```
TJB4epam, line 14: Locked -> Allowed or Disabled?

src/main/java/com/epam/prejap/tetris/block/BlockRotator.java
```diff
@@ -0,0 +1,127 @@
...
+
+    public Block getRotatedBlock() {
+        return block;
+    }
+
+    public Point getBlockOffset() {
+        return blockOffset;
+    }
+
+    public void rotate() {
```
TJB4epam, line 34: Why void? This forces you to have a 'getRotatedBlock' because your method relies on side-effects. :/

src/main/java/com/epam/prejap/tetris/block/BlockRotator.java
```diff
@@ -0,0 +1,127 @@
...
+    public BlockRotator(Block block) {
+        this.block = block;
+        rotationPoint = blockRotationPoint();
+        blockOffset = NO_OFFSET;
+
+        lastRotation = new LastBlockRotation();
+        rotationBackLocked = true;
+    }
+
+    public Block getRotatedBlock() {
```
TJB4epam, line 26: Public getter. :(

src/main/java/com/epam/prejap/tetris/block/BlockRotator.java
```diff
@@ -0,0 +1,127 @@
...
+        return block;
+    }
+
+    public Point getBlockOffset() {
```
TJB4epam, line 30: OOP - objects talk to each other.

Direct state manipulation - objects disembowel themselves and offer their innards to other objects via public getters. Because otherwise, those other objects can't do sh*t.

:(

src/main/java/com/epam/prejap/tetris/block/BlockRotator.java
TJB4epam: Or add Is before Locked: IsLocked.

src/main/java/com/epam/prejap/tetris/block/Point.java
```diff
@@ -0,0 +1,27 @@
+package com.epam.prejap.tetris.block;
+
+import java.util.Objects;
+
+public class Point {
```
TJB4epam, line 5: What about records?

src/main/java/com/epam/prejap/tetris/game/Playfield.java
```diff
@@ -36,12 +39,23 @@ public boolean move(Move move) {
...
+            case UP -> rotateBlock();
         }
         moved = moveDown();
         show();
         return moved;
     }
 
+    private void rotateBlock() {
+        blockRotator.rotate();
+        block = blockRotator.getRotatedBlock();
```
TJB4epam, line 51: :(

TJB4epam (changes requested):
There are two types of programs:

1. smart code with dumb data structures
2. smart data structures and dumb code

You went for number 1 and thus you have lots of code here. But if your Blocks would've had 4 possible images as a list, then rotation would be just calling iterator.next(). No locking on backRotate, no non-trivial calculations, just a list with images in each block. After all, there are only 4 possibilities.

src/main/java/com/epam/prejap/tetris/block/Point.java
krawmaciej: I would need up the project java version to 15, can I do that? Or should I ask everyone working on this project before doing this?
I really didn't know how to go about it so I went with creating a Point class, I think I could also add to it a subtract() and negate() methods too (simpler code in BlockRotator), would it be fine for a record to have them?

src/main/java/com/epam/prejap/tetris/block/BlockRotator.java
krawmaciej: I didn't know how to go about doing this, since this method creates two objects which are used to rotate block in the game:
Block rotatedBlock,
Point howMuchWasTheRotationPointOffsetAfterRotation,
should I create a object for that which is returned by this function (this would also remove all the getters),
and then this object would be used inside a main game to make a block rotation.

src/main/java/com/epam/prejap/tetris/block/BlockRotator.java
krawmaciej: rotation back is disabled, I will change it to isRotationBackLocked

src/main/java/com/epam/prejap/tetris/game/Playfield.java
krawmaciej: Yup I am sorry for that, same reasoning as with rotate(), please see [here](https://github.com/EPAM-Online-Courses/tetris/pull/33#discussion_r554343669)

krawmaciej:
@TJB4epam 
> There are two types of programs:
> 
> 1. smart code with dumb data structures
> 2. smart data structures and dumb code
> 
> You went for number 1 and thus you have lots of code here. But if your Blocks would've had 4 possible images as a list, then rotation would be just calling iterator.next(). No locking on backRotate, no non-trivial calculations, just a list with images in each block. After all, there are only 4 possibilities.

I asked Paweł Kierat if I could add a requirement for the blocks to have square image matrix (and their rotation point in the middle), or even a rotation point held by the Block in a field, to which I got an answer:
> [Wednesday 8:54 PM] Pawel Kierat (Contractor)
> I managed to devise a rotation formula without much effort... AFAIR, you should not have an issue with linear algebra :P

So because of needing to use the rotation formula I couldn't force Block implementers to provide me images for each rotation.

Also because block images aren't required to be squares, then after rotating matrix image, e.g. from 2x3 to 3x2, the rotation point also rotates (isn't kept in the middle), so I needed to keep the offset of rotation point after the rotation, which is then used by main program to move the block (so the rotation point is again where it was before the rotation).

That's why my rotate() method needs to return two values, rotated block and a middle point offset.

src/main/java/com/epam/prejap/tetris/block/BlockRotator.java
krawmaciej: Okay I think I have an idea how to resolve all those public getters, let me implement it and push changes :)

src/main/java/com/epam/prejap/tetris/block/IBlock.java
```diff
@@ -22,4 +22,12 @@
         super(IMAGE);
     }
 
+    public IBlock(byte[][] image) {
```
pkierat-epam, line 25: Why is this constructor public, considering that in other classes it's much better encapsulated?

src/main/java/com/epam/prejap/tetris/block/Point.java
pkierat-epam: AFAIR, records were added as a preview feature in Java 14, and are still in preview in Java 15, so it doesn't seem to matter which of those two versions is in the project. Leaving 14 doesn't introduce conflicts though.

pkierat-epam:
@krawmaciej

1. The project was called "Tetris" for simplicity, but, as the README file states, it's a *Tetris-like* game. Hence, contrary to your assumptions, it doesn't have to be 1-1 clone (and it was never meant to be).
1. AFAIR, you asked about changing the matrix into a square, which doesn't really simplify the calculations too much, since at least the "I" block is 4x4. Moreover, that shouldn't stop you in any way from adding the rotated images in a hard-coded form, should you need it.
1. As for the formula, I only wrote about it since you never even mentioned that you would like to hard-code the images (again, an assumption).
1. Rotation images you proposed in the issue comment suggested that you actually want to rotate the block around the point which is anchored in the playfield. Why not simply rotate it first, and then let the playfield render it using exactly the same point (which is always (0,0))? Would it change the result that much?
   ```
   #     ###   ##      #
   #  -> #  ->  # -> ###
   ##           # 
   ```
Even if you wanted to write a universal rotation algorithm for every kind of block, it's much easier when you rotate it without considering the rotation point at all, and dimensions of the matrix don't really matter here (composing transposition with reflection should be just enough to achieve the desired effect).

src/main/java/com/epam/prejap/tetris/block/Point.java
krawmaciej: @pkierat-epam 
Thank you! So this is how trusting your IDE with everything ends...
It showed me an error and to resolve I needed to change java version to 15,
automatically inserted code:
```
<source>15</source>
<target>15</target>
<compilerArgs>--enable-preview</compilerArgs>
```
and after leaving only ` <compilerArgs>--enable-preview</compilerArgs> ` so it uses source and target from properties (set to java 14) it of course works too.

For the code above record is perfect solution, I still have a question out of curiosity:
If I were to add custom methods to Point class, e.g. Point subtract(Point), Point negate(), Point add(Point), then is letting it stay as a class a good idea or should it be also converted to record? (record gives good chunk of functionality for free so I would only need to add those 3 methods and nothing more, that is what brought my question)

src/main/java/com/epam/prejap/tetris/block/IBlock.java
krawmaciej: @pkierat-epam 
IDE autogenerated constructor, I changed it for other classes but forgot for this one, fortunately for me this whole class is package-private. However I will need to check if I can change access modifiers of autogenerated methods in IDE to avoid it in the future.
Thanks for pointing that out!

pkierat-epam:
Records support custom methods, so I see only benefits :-)
________________________________
Od: Maciej Krawczyk <notifications@github.com>
Wysłano: niedziela, 10 stycznia 2021 15:01
Do: EPAM-Online-Courses/tetris
DW: Pawel Kierat (Contractor); Mention
Temat: Re: [EPAM-Online-Courses/tetris] 7 Rotation of blocks (#33)


@krawmaciej commented on this pull request.

________________________________

In src/main/java/com/epam/prejap/tetris/block/Point.java<https://urldefense.com/v3/__https://github.com/EPAM-Online-Courses/tetris/pull/33*discussion_r554571820__;Iw!!GF_29dbcQIUBPA!n5OVOHI9XpiEYNuvw2KppucUP78udMbLSYz8-P8BZTkuOJoL6k8jEJLREOq5W8NOyA$>:

> @@ -0,0 +1,27 @@
+
+import java.util.Objects;
+
+public class Point {


@pkierat-epam<https://urldefense.com/v3/__https://github.com/pkierat-epam__;!!GF_29dbcQIUBPA!n5OVOHI9XpiEYNuvw2KppucUP78udMbLSYz8-P8BZTkuOJoL6k8jEJLREOqtWw1ZPQ$>
Thank you! So this is how trusting your IDE with everything ends...
It showed me an error and to resolve I needed to change java version to 15,
automatically inserted code:

<source>15</source>
<target>15</target>
<compilerArgs>--enable-preview</compilerArgs>


and after leaving only <compilerArgs>--enable-preview</compilerArgs> so it uses source and target from properties (set to java 14) it of course works too.

For the code above record is perfect solution, I still have a question out of curiosity:
If I were to add custom methods to Point class, e.g. Point subtract(Point), Point negate(), Point add(Point), then is letting it stay as a class a good idea or should it be also converted to record? (record gives good chunk of functionality for free so I would only need to add those 3 methods and nothing more, that is what brought my question)

krawmaciej:
@pkierat-epam
Hi! there goes a wall of text as a reply to [this message](https://github.com/EPAM-Online-Courses/tetris/pull/33#issuecomment-757447171)

1. At first I wanted to implement Tetris rotations, but without rotation point set for each block nor having square matrices it would be very hard, therefore I went for the closest but still kinda resembling original Tetris rotations, which also had its own problems (mainly rotation point moving on the matrix).
I took it as a challenge to try to implement that using existing program functionality, so it's all on me and there are no excuses.

Therefore there are no more excuses for other points, I will remember to ask more questions and clarify everything that I am going to do as I am doing it.


However, thanks to implementing it this way I have many questions :) :

### Method returning value and modifying object's state:
My code depends on two results from a single rotation() action, and this action also modifies the object's state.

Let me explain:
I have a BlockRotator object which hold a Block that it's instantiated with and its rotation point.

After calling rotate() method it updates its state, now it holds rotated block and rotated rotation point, I need both these results in main game loop, to update current block and offset render point by how much the rotation point was offset.

Because this method has a side effect it should not return a value, so I decided to cheat around that rule and get this result from public getters.

I know of some classes' methods that also break this rule because they need to, and I want to know your point of view on that, how should I go about implementing a method which has side effect but also returns a result?
e.g. classes breaking this rule would be
Stack.pop(), Queue.remove()/poll(), Iterator.next(),

However, the above do not share the state that is modified in the side effect outside, so would returning copies of those modified state objects from rotate() be fine?

### What can be modified in existing code:

My solution also came from the fact that I wanted to keep modification of existing code as low as possible, I also wrongly assumed that from your answer (bad I know, should have asked when I saw the need to modify existing code).
I understood it that maybe forcing other implementers to follow some requirements of creating a block image, or adding a rotation point to Block class and forcing implementers to provide it in their Block code was a bad approach of solving this - too many existing code changes.


What changes to existing code can be done, and what shouldn't? I feel like answer to that is "it depends", but maybe there are some kind of rules, or will it all come with experience?

#### Different solutions to a rotating block
For example would changes described below be fine (or am I still overthinking it, and of course they would be fine):

1. Let's say that I wanted to add to a Block parent class a private final rotation point field (and also a rotate() method, because they are connected, and because rotate is what a block does), this field would be instantiated inside a Block constructor, and then after rotation it would be changed appropriately.
Still I need give this rotationPoint outside of the class, same idea as with:
```
public byte dotAt(int i, int j) {
    return image[i][j];
}
```
so since public getters are a no-no, should I cheat and add method named public Point rotationPoint() returning a copy of rotation point outside? or would even a public getter returning a copy be fine?
The above would result in change to only one class and doesn't force anything on concrete implementations.


2. There is also a other idea of doing this, brought to me by @LukasBulczak, which I feel is better, because it creates a RotatableBlock interface, implemented by RotatingBlock. Interface RotatableBlock is a wrapper for Block, which adds to it a rotate functionality.
To use this functionality in main game, private Block block field type in Playfield class needs to be changed to RotatableBlock, which would not introduce any more changes to the existing code.
If someone wishes to use instanceof on this RotatableBlock, to check if it's actually a LBlock or other underneath, then RotatableBlock also proves a method asBlock() which would return a copy of the block it wraps, and on that block an instanceof can be used.


Are those changes fine, or maybe you know some situations where wrapping an object in existing codebase introduced many problems and there are better solutions?


3. Also another approach would be of doing the same as above, but using inheritance. Create an abstract RotatingBlock class (not an interface because it needs to hold rotatingPoint field). Make Block extend that class, and all the other stuff stays pretty much the same, but we don't need to use asBlock() method to check instanceof.

Which approach would be better (I already feel like the last one is worst because of inheritance), what are their advantages and disadvantages?

krawmaciej:
I added new commits, resolving the issues mentioned above, I choose the approach that I talked with @LucasBulczak about (I wrote about it [here](https://github.com/EPAM-Online-Courses/tetris/pull/33#issuecomment-757545998)).

The changes I had to make in Playfield class, which I want to ask about were as following:

1. changed and renamed field from `private Block block;` to `private RotatableBlock rotatableBlock;` (line 14):
should I leave the name as block, because now every reference to block was renamed to rotatableBlock?

2. changed first argument type of method `private boolean isValidMove(Block rotatableBlock` from Block to RotatableBlock (line 93)
should I leave it as it was (so Block type) and unwrap this object before calling this method, which would result in change in the line 86:
from
`if (isValidMove(rotatableBlock, rowOffset, colOffset))`
to
`if (isValidMove(rotatableBlock.asBlock(), rowOffset, colOffset))`

krawmaciej:
I added tests for block rotations using factory, which generates rotation tests for all blocks, testing those blocks against expected rotation images and expected rotation points.
I added those rotation images and rotation points as static fields in a package-private class BlockRotations in test directory, is that a good approach for providing test data for a factory?

krawmaciej:
@pkierat-epam @TJB4epam
Hi guys!
After adding the tests for block rotations - both their "dot" image structure and in game (in game not for every block though :D), I think the PR is ready for the code review :).

To summarize changes (with questions that came up):
I created an interface RotatableBlock which wraps a Block object, this interface is implemented by RotatingBlock, which in turn is returned by interface static .of(Block) method.
This wrapper doubles on the Block functionality so it can be used in place of Block object, and its implementation delegates those functions to to the Block object it wraps around.
rotate() method now returns new rotated object, so no more side effects and using getters to get the state modified by side effects :)
I also added JD explaining what those methods do.

Questions:
- ArrayPoint rotationPoint() returns object held in private rotationPoint method, without any defensive copying, however ArrayPoint is a record holding two ints, so it is effectively final, can it stay like that, or should I return a copy.
- same thing for asBlock() method, which returns a Block that this interface wraps around.
Should I create a copy of wrapped Block at instantiation and on returning it, or since interface wraps this object (giving it additional functionality), then maybe it's fine for it to refer to original object? or is it never fine to not create a copy?

Tests:
I created two factories (BlockRotationTestFactory, PlayfieldBlockRotationTestFactory) for checking block rotation:
1. for BlockRotationTest (testing block objects) I pass block rotation images and block rotation points (for 0°, 90°, 180°, 270°, 360° rotations) from [record](https://github.com/EPAM-Online-Courses/tetris/blob/e3947db7396cf6ff0636796bf4fe918ef3a720eb/src/test/java/com/epam/prejap/tetris/block/BlockRotations.java), and I use DataProvider inside BlockRotationTest which uses oridinal values from RotationDegree enum, indicating how many times a block has to be rotated.
2. for BlockRotationTest (testing block rotations on a playfield's grid) I needed to create a mock of a BlockFeed, I came up with the same idea as Paweł Kwieciński had, to override BlockFeed inside test block package and by making it public I can access this mock from my test game package.

Question (this is the question I wanted to ask during today's lecture): I didn't want to make the mock dependent on indices on which the blocks are held inside BlockFeed, so I don't use random or blocks List, instead I create a new block that I need with a Supplier.
also I still couldn't pass Blocks nor their methods to this mocked blockfeed (because block classes are package-private) so I created
```
public static BlockFeed iBlockFeed() {
    return new MockBlockFeed(IBlock::new);
}
```
methods for each block, and I use that on my playfield test, is it a good approach to solving this issue? otherwise I think I needed to make block classes and their constructors public, or create some gateway class in block package which returned needed instances of blocks (so kinda what I am doing in MockBlockFeed right now).

Also, is making parametrized test for each next block rotation a good idea, or should I create a distinct test for each rotation (I did it this way at first, but there was a lot of code duplication) with name clearly explaining what should have happened after what actions?
Is it possible to make a name of a test depending on one of the given parameters from DataProvider?

Cheers!

TJB4epam:
Further new changes merged. Also, please make sure to reply to previous comments. Your promised answer is incoming - should be today (MSTeams).

krawmaciej:
@TJB4epam @pkierat-epam 
Okay I've merged the changes from master branch, I had to change "some" stuff :D, most time took me debugging of a PlayfieldTest tests because sometimes they passed and sometimes they didn't.
It used all possible Moves from Move enum, and after my changes it also used rotation, which wasn't taken into account for this test so I decided to change it to use only the moves it was designed for.

src/main/java/com/epam/prejap/tetris/block/Block.java
```diff
@@ -1,5 +1,7 @@
 package com.epam.prejap.tetris.block;
 
+import java.util.HashMap;
```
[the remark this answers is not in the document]
krawmaciej: I've changed my IDE settings to automatically optimize imports of a file I save.

src/main/java/com/epam/prejap/tetris/block/BlockRotator.java
krawmaciej: Now it's entirely removed, as there is no need for that flag or even throwing an exception.

src/main/java/com/epam/prejap/tetris/block/BlockRotator.java
krawmaciej: Done - not needed now. I use something similar but different now :D, I asked question about it [here](https://github.com/EPAM-Online-Courses/tetris/pull/33#issuecomment-759488485).

krawmaciej:
@pkierat-epam 
Concerning [this](https://github.com/EPAM-Online-Courses/tetris/pull/39#discussion_r557088524), which @NikaAvramchuk mentioned in a post on Teams.
Because I wrote classes like that, which are used by the Tests, should I rename them from e.g. BlockRotations to TestBlockRotations ?
also what about [mock block feed](https://github.com/EPAM-Online-Courses/tetris/blob/adad40d10405c3a17528dc000e0ef69f3430219f/src/test/java/com/epam/prejap/tetris/block/MockBlockFeed.java) class I created, it's also used only by the tests, should it be renamed to TestMockBlockFeed? or because it has a Mock in its name there is no need for that?

TJB4epam:
> @TJB4epam @pkierat-epam
> Okay I've merged the changes from master branch, I had to change "some" stuff :D, most time took me debugging of a PlayfieldTest tests because sometimes they passed and sometimes they didn't.
> It used all possible Moves from Move enum, and after my changes it also used rotation, which wasn't taken into account for this test so I decided to change it to use only the moves it was designed for.

That's OK if you have tests that focus on rotation.

TJB4epam:
> @pkierat-epam
> Concerning [this](https://github.com/EPAM-Online-Courses/tetris/pull/39#discussion_r557088524), which @NikaAvramchuk mentioned in a post on Teams.
> Because I wrote classes like that, which are used by the Tests, should I rename them from e.g. BlockRotations to TestBlockRotations ?
> also what about [mock block feed](https://github.com/EPAM-Online-Courses/tetris/blob/adad40d10405c3a17528dc000e0ef69f3430219f/src/test/java/com/epam/prejap/tetris/block/MockBlockFeed.java) class I created, it's also used only by the tests, should it be renamed to TestMockBlockFeed? or because it has a Mock in its name there is no need for that?

Naming convention is for TEST CLASSES. Not HELPER test classes. 
Naming convention is for said classes to be found by default Maven execution, without testng.xml file.

But, if a project adopts a naming convention, do you have the time to take care of test names of all tests and to present a convincing argument that your tests are exceptional enough to yada yada yada? Today you don't as you still have conflicts. So change test classes as per convention.

krawmaciej:
> > @pkierat-epam
> > Concerning [this](https://github.com/EPAM-Online-Courses/tetris/pull/39#discussion_r557088524), which @NikaAvramchuk mentioned in a post on Teams.
> > Because I wrote classes like that, which are used by the Tests, should I rename them from e.g. BlockRotations to TestBlockRotations ?
> > also what about [mock block feed](https://github.com/EPAM-Online-Courses/tetris/blob/adad40d10405c3a17528dc000e0ef69f3430219f/src/test/java/com/epam/prejap/tetris/block/MockBlockFeed.java) class I created, it's also used only by the tests, should it be renamed to TestMockBlockFeed? or because it has a Mock in its name there is no need for that?
> 
> Naming convention is for TEST CLASSES. Not HELPER test classes.
> Naming convention is for said classes to be found by default Maven execution, without testng.xml file.
> 
> But, if a project adopts a naming convention, do you have the time to take care of test names of all tests and to present a convincing argument that your tests are exceptional enough to yada yada yada? Today you don't as you still have conflicts. So change test classes as per convention.

Okay I've misunderstood what @pkierat-epam wrote, I will put the unclear part here:
> The difference between between SomethingTest and TestSomething is that the former is a test for something, and the latter is a something used for tests.

I thought there is also a convention of naming test helper classes, with adding Test before the name of the helper class.
Even if there is not, I saw that Veronika already used that notation for her classes used in tests, e.g. **Test**IBlock, **Test**JBlock, which are used by BlockTestFactory. So to not cause any confusion I am gonna follow the same naming for helper classes.

krawmaciej:
@TJB4epam 
Hi! Everything merged, tests for rotating every block added. I went with the approach you told me about, to make blocks implement Iterable and use iterator on block rotation images to get next block rotation.

src/test/java/com/epam/prejap/tetris/game/PlayfieldBlockRotationTest.java
```diff
@@ -0,0 +1,77 @@
...
+import org.testng.annotations.Test;
+
+import java.io.ByteArrayOutputStream;
+import java.io.PrintStream;
+import java.util.List;
+import java.util.regex.Pattern;
+
+import static org.testng.Assert.assertTrue;
+
+@Test(groups = "Playfield", dependsOnGroups = "BlockRotation")
```
TJB4epam, line 15: Good idea.

TJB4epam (approved):

pom.xml
```diff
@@ -42,6 +42,16 @@
         <finalName>${project.artifactId}</finalName>
         <plugins>
 
+            <plugin>
```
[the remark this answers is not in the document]
TJB4epam: Resolved a moment later with a commit.

TJB4epam:
63 + 10 = 73 points. Nice.

Issue #41, opened 2021-01-13
Fix the errors when tests are executing


TJB4epam:
It's ERROR / WARNING from Mockito illegal reflective access, excerpt from build logs below:

```
[ERROR] WARNING: An illegal reflective access operation has occurred
[ERROR] WARNING: Illegal reflective access by org.mockito.cglib.core.ReflectUtils$2 (file:/home/tammo/.m2/repository/org/mockito/mockito-all/1.10.19/mockito-all-1.10.19.jar) to method java.lang.ClassLoader.defineClass(java.lang.String,byte[],int,int,java.security.ProtectionDomain)
[ERROR] WARNING: Please consider reporting this to the maintainers of org.mockito.cglib.core.ReflectUtils$2
[ERROR] WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
[ERROR] WARNING: All illegal access operations will be denied in a future release
```

You may want to use mvn test and git blame and see when this was introduced. We need to get rid of it but retain mocking in tests or keep the tests working / fast / isolated. Feel free to engage the original authors who introduced Mockito.

WiktoriaMajchrzak:
I can try to fix it :)

TJB4epam:
Excellent! Deadline is Friday, 15th Jan, 17:00 CET.

TJB4epam:
Done by Wiktoria. Closing.

PR #44, opened 2021-01-13
41 Fix the errors when test are executing

WiktoriaMajchrzak:
Dependency mockito-all has been replaced with depencency mockito-core. The mockito-core artifact is Mockito's main artifact. “Mockito-all” distribution has been discontinued in Mockito 2.*. The latest GA version of mockito-all is a 1.x version released in 2014. Newer versions of Mockito don't release mockito-all anymore.

@pkierat-epam I have created PR again. The previous one had some git history issues.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
     <modelVersion>4.0.0</modelVersion>
     <groupId>com.epam.prejap</groupId>
     <artifactId>tetris</artifactId>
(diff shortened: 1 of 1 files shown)


TJB4epam:
Good solution. Closes the #41 - merging.

TJB4epam:
10 points awarded.

PR #50, opened 2021-01-15
12 every block has different color - Fix bug in testng.xml

dominikjaniga91:
Hi @TJB4epam, in testng.xml there was Color class instead of ColorTest, I fixed that.


pom.xml
```diff
@@ -89,9 +89,9 @@
                 <artifactId>maven-compiler-plugin</artifactId>
                 <version>3.8.1</version>
                 <configuration>
-                    <compilerArgs>
-                        <arg>--enable-preview</arg>
-                    </compilerArgs>
+                    <compilerArgs>--enable-preview</compilerArgs>
+                    <source>14</source>
```
TJB4epam, line 93: 1. Wasn't the source and target set via properties already?
2. Why so many commits for just two files? Would you kindly base this off master?

src/test/resources/testng.xml
```diff
@@ -51,7 +51,7 @@
             <class name="com.epam.prejap.tetris.game.PlayfieldTest"/>
             <class name="com.epam.prejap.tetris.game.TimerTest"/>
             <class name="com.epam.prejap.tetris.game.RefereeTest"/>
-            <class name="com.epam.prejap.tetris.block.Color"/>
+            <class name="com.epam.prejap.tetris.block.ColorTest"/>
```
TJB4epam, line 54: 'preciated.

TJB4epam (changes requested):

PR #51, opened 2021-01-15
Fix bug in testng xml

dominikjaniga91:
Hi @TJB4epam, I removed previous PR and created new one from fix-branch.

1. Bug in testng.xml fixed
2. I moved all versions in pom.xml to properties

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -16,9 +16,11 @@
         <version.plugin.maven.jar>3.2.0</version.plugin.maven.jar>
         <version.plugin.maven.surefire>3.0.0-M5</version.plugin.maven.surefire>
         <version.plugin.maven.failsafe>3.0.0-M5</version.plugin.maven.failsafe>
+        <version.plugin.maven.compiler>3.8.1</version.plugin.maven.compiler>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -87,18 +89,18 @@
...
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-compiler-plugin</artifactId>
-                <version>3.8.1</version>
+                <version>${version.plugin.maven.compiler}</version>
                 <configuration>
-                    <compilerArgs>
-                        <arg>--enable-preview</arg>
-                    </compilerArgs>
+                    <compilerArgs>--enable-preview</compilerArgs>
+                    <source>${project.java.version}</source>
```
TJB4epam, line 95: OK, what about line 13 and 14 in the POM?

EDIT: correct numbers

dominikjaniga91: Alright, I removed those lines

TJB4epam (approved):

TJB4epam:
10 points for fix. Re tests: will see if they were counted before or not. There will be no counting twice. ;-)

Issue #19, opened 2021-01-05
The app doesn't run with 'java -jar'


pkierat-epam:
## Description

When trying to run the app using an executable jar I get ClassNotFoundException.

## How to reproduce

```
$ mvn clean package
[...]
$ java -jar target/tetris-0.1.jar 
Error: Could not find or load main class com.epam.prejap.sdp.git.tetris.Tetris
Caused by: java.lang.ClassNotFoundException: com.epam.prejap.sdp.git.tetris.Tetris
```

## Expected behavior

```
$ mvn clean package
[...]
$ java -jar target/tetris-0.1.jar
+--------------------+
|         ##         |
|         ##         |
|                    |
|                    |
|                    |
|                    |
|                    |
|                    |
|                    |
|                    |
+--------------------+
[...]

Szymon-Skazinski:
I'm working on this task

TJB4epam:
Deadline is Friday, 15th Jan, 17:00 CET.

AndriiVel:
I'm working on this task.

PR #65, opened 2021-06-06
19 run jar from console

AndriiVel:
This PR fixes issue #19  where app could not be run using an executable JAR.

**Changes:**

- In `pom.xml` the manifest section in `maven-jar-plugin` was corrected.
- Added tests to check whether main class is specified in JAR manifest.
- Added new dependency on `properties-maven-plugin` to update data for tests, in case project name or version is changed.


pom.xml
```diff
@@ -6,19 +6,30 @@
...
     <groupId>com.epam.prejap</groupId>
     <artifactId>tetris</artifactId>
-    <version>0.1</version>
+    <version>0.2</version>
 
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <maven.compiler.source>16</maven.compiler.source>
         <maven.compiler.target>16</maven.compiler.target>
+        <testng.version>7.4.0</testng.version>
```
TJB4epam, line 15: I love moving everything to props. Kudos.

To make it better: let's have a convention for VERSIONS. So, Everything that's a version should be done similarly to testng.version. That way one may not think what the version was: just name.version will do.

pom.xml
```diff
@@ -28,17 +39,53 @@
...
                             <addClasspath>true</addClasspath>
-                            <mainClass>com.epam.prejap.git.tetris.Tetris</mainClass>
+                            <mainClass>${main.class}</mainClass>
                         </manifest>
                     </archive>
                 </configuration>
+                <executions>
+                    <execution>
+                        <id>jar-creation-for-testing</id>
+                        <phase>generate-test-resources</phase>
```
TJB4epam, line 54: Creative and nicely helps with testing being before packaging. :)

Do remember to make this used as an integration test!

pom.xml
```diff
@@ -28,17 +39,53 @@
...
+                            <classifier>${test.jar.name}</classifier>
+                        </configuration>
+                        <goals>
+                            <goal>jar</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
+
+            <plugin>
```
TJB4epam, line 72: OK, I see why this plug-in but I doubt that without a line of commentary others will know that as well. I'd like for you to write short doc in the test file, as a Javadoc as put a comment here saying "see this test file for why we need this". That way, in future we will be able to review the decision and change it if necessary.

src/test/java/com/epam/prejap/tetris/JARTest.java
```diff
@@ -0,0 +1,69 @@
...
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Properties;
+import java.util.jar.JarFile;
+
+import static org.testng.Assert.*;
+
+@Test(groups = "jar")
```
TJB4epam, line 14: OK - I want this to run only with integration tests, as these are not unit tests. I want `mvn test` to NOT run this test but for this test to run when we do `mvn package`.

TJB4epam (commented):
Integration tests should be made as such. A bit docs need to be written. Nice job otherwise!

pom.xml
AndriiVel: Done. I also moved the versions to the bottom of `<properties>` to make it more readable in case we want to add versions of new dependencies.

pom.xml
AndriiVel: I'm not sure whether I'll need it in case I run my tests as integration tests on jar in `target/`.

This execution generates a duplicated jar and indeed helps with testing before packaging :)

pom.xml
AndriiVel: Good idea. Done.

src/test/java/com/epam/prejap/tetris/JARTest.java
AndriiVel: Done.

src/test/java/com/epam/prejap/tetris/JARArchiveIT.java
```diff
@@ -0,0 +1,74 @@
...
+
+import static org.testng.Assert.*;
+
+/**
+ * Integration tests that check whether JAR archive was correctly created.
+ * The class gets correct JAR file name from properties extracted from <b>pom.xml</b> with
+ * <b>properties-maven-plugin</b>.
+ */
+@Test(groups = "jar")
+public class JARArchiveIT {
```
AndriiVel, line 20: I renamed the class, cause by default maven-failsafe-plugin will run all classes that starts with "IT" or ends with "IT" or "ITCase".

marcin-ogorzalek-EPAM:
@AndriiVel please resolve merge conflicts

AndriiVel:
Merge conflicts resolved.

pom.xml
```diff
@@ -6,21 +6,26 @@
...
         <maven.compiler.source>16</maven.compiler.source>
         <maven.compiler.target>16</maven.compiler.target>
+        <main.class>com.epam.prejap.tetris.Tetris</main.class>
+        <test.jar.file.location>
+            ${project.build.directory}${file.separator}${project.build.finalName}.jar
+        </test.jar.file.location>
+        <testng.version>7.4.0</testng.version>
         <maven.jar.plugin.version>3.2.0</maven.jar.plugin.version>
-        <testing>7.4.0</testing>
+        <properties.maven.plugin.version>1.0.0</properties.maven.plugin.version>
```
TJB4epam, line 21: Plugin convention doesn't fit. Minor, but important. Maven JAR plugin version vs Properties Maven plugin version. See one line above (not counting the cut line).

pom.xml
```diff
@@ -35,12 +40,56 @@
...
                     </archive>
                 </configuration>
             </plugin>
+
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <version>3.0.0-M5</version>
+                <configuration>
+                    <excludedGroups>jar</excludedGroups>
```
TJB4epam, line 54: Nice.

src/test/java/com/epam/prejap/tetris/JARArchiveIT.java
TJB4epam: Rightly so.

src/test/resources/testng.xml
```diff
@@ -1,14 +1,17 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!DOCTYPE suite SYSTEM "http://testng.org/testng-1.0.dtd">
+<!DOCTYPE suite SYSTEM "https://testng.org/testng-1.0.dtd">
```
TJB4epam, line 2: VERY good. Otherwise things would went bad pretty fast. :)

src/test/resources/testng.xml
```diff
@@ -1,14 +1,17 @@
...
+
 <suite name="TestNG Group Test">
     <test verbose="2" name="Tetris tests">
         <groups>
             <run>
                 <include name="blockShapes"/>
+                <include name="jar"/>
```
TJB4epam, line 9: Hmmm... now running tests for TestNG we are running both unit and integration tests, are we not?

How about creating a super group to separate them?

TJB4epam (approved):
I am approving / merging this cause this stops others, but the comments I left should be tackled later on. FYI, @AndriiVel

src/test/resources/testng.xml
AndriiVel: Will do. Shall I create a new issue and new PR for it?

pom.xml
TJB4epam: Adding a comment illustrating this will suffice IMO.

src/test/resources/testng.xml
TJB4epam: Definitely.

Issue #67, opened 2021-06-07
Add welcome screen on launch


rpiw:
1. User story: as a user I would like to see welcome string when launching program
2. Acceptance criteria: String is visible for at least 3 seconds
3. Additional points for adding a possibility to interrupt welcome screen with a pressing any key.

Winetq:
I'm working on this task.

rpiw:
@Winetq if you want to make it even fancier, try to add user name (from machine)

Winetq:
great ideas, I will try and thank you very much for the issue :)

PR #76, opened 2021-06-10
67 add welcome screen on launch

Winetq:
I added a welcome screen as a thread. It is visible for given number of seconds, but there is a possibility to manually close the screen pressing the Esc key. Of course I also provided a test to check this functionality.

To sum up: solve the #67 issue


Adam-Szczeblewski-EPAM:
JDK has a built-in splashscreen capability, is it worth to roll out a custom solution?
https://docs.oracle.com/javase/tutorial/uiswing/misc/splashscreen.html

Also, your test will fail in most CI environments, as they are headless, and won't allow creating any frames.
Possible solutions:
- provide a CI runner which is not headless
- exclude the test from running in headless environment. in junit [@DisabledIf...](https://junit.org/junit5/docs/5.2.0/api/org/junit/jupiter/api/condition/DisabledIfSystemProperty.html)  could be used. In TestNG - I'm sure there's a similar way

Winetq:
To be honest I would need more time to implement a splash screen, because I'm struggling with returning null from the getSplashScreen() method. For now I think that my solution is sufficient. BTW I can't remind myself what is the CI runner.

Adam-Szczeblewski-EPAM:
CI - ContinuoutIntegration. E.g. Github Checks, to executed tests upon git push.

I see there are no Checks in this repo, so no CI is configured at all. So for now - yes, your solution is sufficient.
Might need to exclude this test (as in my bullet point 2) once CI is set up.

TJB4epam:
Conflict reminder.

TJB4epam (changes requested):
I LOVE the idea. However, I'd like several changes here.

1. There's no mention about splash screen in this, and I'd like for the docs to say so - that there is such option and that we would like to rework this. Splash screen is pretty much what we're implementing here.
2. Tests for the greeting do not mention that they won't work in a headless env. I'd like for that to be more prominent: group name should reflect it.
3. All-headless-tests group should be done, and greeting tests should NOT be a part of it. 
4. Separate testng.xml file is a nice touch, locating it in test/resources is a nice touch as well, but why did you do it? I can think of a few reasons, people not involved in this PR will be confused. Let's remedy that: rename that file in a meaningful way. Put a comment inside about why it's needed.
5. I can't let it in without bumping the version in the POM.
6. Javadocs should NOT start with "This javadocced-element" because we know that we're reading a class JD or a method one. So reading "This method" on a method is pointless. Kill all "This method" or "This class" or "Name of class class" comment elements - mind you, elements, not entire comments that have them.
7. If you're not gonna use Adam's comments w.r.t. splash screen, add a TODO comment and a new issue. TODO: link to the new issue, in the issue, write that splash screen could be redone and add Adam's comments about it.
8. You are changing the manifest! Sensible... but what about changing the JAR tests that test said manifest? New changes, by @AndriiVel, PR #19.

pom.xml
```diff
@@ -6,7 +6,7 @@
 
     <groupId>com.epam.prejap</groupId>
     <artifactId>tetris</artifactId>
-    <version>0.4</version>
+    <version>0.5</version>
```
TJB4epam, line 9: TY!

pom.xml
```diff
@@ -17,6 +17,7 @@
             ${project.build.directory}${file.separator}${project.build.finalName}.jar
         </test.jar.file.location>
         <testng.version>7.4.0</testng.version>
+        <version.assertj>3.19.0</version.assertj>
```
TJB4epam, line 20: :+1:

pom.xml
```diff
@@ -28,6 +29,17 @@
...
+        <dependency>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
+            <version>${version.assertj}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-failsafe-plugin</artifactId>
+            <version>3.0.0-M5</version>
```
TJB4epam, line 41: Put the version in the properties please.

Also, I have a feeling we already have this?

TJB4epam: Checked and indeed we do. Take a look why would it be marked as new in your case. Perhaps git rebase might be in order.

Version still should go to properties, the merged code has it directly on the plug-in too.

src/main/java/com/epam/prejap/tetris/Tetris.java
```diff
@@ -51,6 +49,7 @@ public static void main(String[] args) {
         var playfield = new Playfield(rows, cols, feed, printer);
         var game = new Tetris(playfield, new Waiter(delay), new RandomPlayer());
 
+        new Greeting(10000).welcomeThePlayer();
```
TJB4epam, line 52: 1. Number literals are better written this way: `10_000` - clearer to read.
2. Why this particular magic number? Why not... 10_020?

src/main/java/com/epam/prejap/tetris/greeting/Greeting.java
```diff
@@ -0,0 +1,35 @@
+package com.epam.prejap.tetris.greeting;
+
+/**
+ * The Greeting class manage the Welcomer thread.
```
TJB4epam, line 4: Javadocs should NOT start with "this is the Javadoc for name-of-element element".

This class
This method
This interface
This enum
...
...
...


NO

TJB4epam: Please change "the Welcomer thread" into a link. Use link, linkplain, or something equivalent.

Alternatively: a `see` tag. 

Also, I'm not sure about a thread being in it's own class, but we shall see when I get there. :-)

src/main/java/com/epam/prejap/tetris/greeting/Greeting.java
```diff
@@ -0,0 +1,35 @@
...
+ * @author Michał Cwynar
+ */
+public class Greeting {
+    private final Thread welcomer;
+
+    /**
+     * The constructor creates the thread.
```
TJB4epam, line 11: The constructor... part has to go.

An improvement for you to consider for testability: instead of having a c-tor that creates a thread, perhaps some constructional design pattern? Static factory method? Builder?

src/main/java/com/epam/prejap/tetris/greeting/Welcomer.java
```diff
@@ -0,0 +1,73 @@
...
+import javax.swing.JFrame;
+import javax.swing.JLabel;
+import javax.swing.JPanel;
+import javax.swing.WindowConstants;
+import java.awt.BorderLayout;
+import java.awt.event.KeyEvent;
+import java.awt.event.KeyListener;
+
+/**
+ * This thread has one task to do which is welcoming the player.
```
TJB4epam, line 12: Smart opening choice. I approve!

Good JD, here.

src/main/java/com/epam/prejap/tetris/greeting/Welcomer.java
```diff
@@ -0,0 +1,73 @@
...
+    /**
+     * The task of this thread is to create a welcome screen using Java Swing and display it for given number of
+     * milliseconds and after that deleting this screen.
+     */
+    @Override
+    public void run() {
+        createTheWelcomeWindow();
+        try {
+            Thread.sleep(millis);
+        } catch (InterruptedException ignore) {}
```
TJB4epam, line 33: Serious bad practice. Please, don't do this. https://www.nurkiewicz.com/2014/05/interruptedexception-and-interrupting.html

src/main/java/com/epam/prejap/tetris/greeting/Welcomer.java
```diff
@@ -0,0 +1,73 @@
...
+        dispose();
+    }
+
+    private void createTheWelcomeWindow() {
+        setLocationRelativeTo(null);
+        setTitle("Tetris");
+        setSize(325, 150);
+        setVisible(true);
+        setDefaultCloseOperation(WindowConstants.DO_NOTHING_ON_CLOSE);
```
TJB4epam, line 42: Are we sure? This kinda feel like malpractice, if I launched the program by mistake, why wouldn't I want to close it?

Then again, the story didn't specify it. So, I leave the end decision in your hands, but in the splash screen user story please note that we should think about it.

src/main/java/com/epam/prejap/tetris/greeting/Welcomer.java
```diff
@@ -0,0 +1,73 @@
...
+        setResizable(true);
+        addKeyListener(this);
+        setFocusable(true);
+        setFocusTraversalKeysEnabled(false);
```
TJB4epam, line 46: IIRC this disables tab moving focus during the screen's existence. Why we need this for a splash screen?

src/main/java/com/epam/prejap/tetris/greeting/Welcomer.java
```diff
@@ -0,0 +1,73 @@
...
+        createTheLabels();
+    }
+
+    private void createTheLabels() {
```
TJB4epam, line 50: If we return the panel created we can use this code later. Due to splash screen issue - I leave that to you to decide, you can forego this comment.

src/main/java/com/epam/prejap/tetris/greeting/Welcomer.java
```diff
@@ -0,0 +1,73 @@
...
+        JPanel panel = new JPanel();
+        panel.add(new JLabel("Welcome " + System.getProperty("user.name") + "!"), BorderLayout.CENTER);
+        panel.add(new JLabel("Good luck!"), BorderLayout.CENTER);
+        add(panel);
+    }
+
+    @Override
+    public void keyTyped(KeyEvent e) {}
```
TJB4epam, line 58: If this doesn't deserve a Javadoc comment illuminating others why that's empty...


then this PR doesn't deserve my review! :P

src/main/java/com/epam/prejap/tetris/greeting/Welcomer.java
```diff
@@ -0,0 +1,73 @@
...
+
+    /**
+     * There is a possibility to manually close the screen pressing the Esc key.
```
TJB4epam, line 61: Thanks for the explanation. Please make this a part of program documentation. A sentence like: "you can close the welcome screen with Esc key" will suffice.

src/main/java/com/epam/prejap/tetris/greeting/Welcomer.java
```diff
@@ -0,0 +1,73 @@
...
+     */
+    @Override
+    public void keyPressed(KeyEvent e) {
+        if (e.getKeyCode() == KeyEvent.VK_ESCAPE) {
+            greeting.finishTheWelcoming();
+        }
+    }
+
+    @Override
+    public void keyReleased(KeyEvent e) {}
```
TJB4epam, line 72: See above. Same thing.

src/main/java/com/epam/prejap/tetris/greeting/Greeting.java
```diff
@@ -0,0 +1,35 @@
...
+     */
+    public Greeting(long milis) {
+        this.welcomer = new Thread(new Welcomer(this, milis));
```
TJB4epam, line 14: Just FYI: you are allowing for this reference to escape the c-tor PRIOR to c-tor being done. Serious malpractice w.r.t. multithreading. JCIP (Java Concurrency In Practice) is a recommended reading on the matter.

src/main/java/com/epam/prejap/tetris/greeting/Welcomer.java
```diff
@@ -0,0 +1,73 @@
...
+    public void keyTyped(KeyEvent e) {}
+
+    /**
+     * There is a possibility to manually close the screen pressing the Esc key.
+     * @param e the pressed key
+     */
+    @Override
+    public void keyPressed(KeyEvent e) {
+        if (e.getKeyCode() == KeyEvent.VK_ESCAPE) {
+            greeting.finishTheWelcoming();
```
TJB4epam, line 67: So, if we just rework this, we can have the thread complete normally. A thing worth having!

src/main/java/com/epam/prejap/tetris/greeting/Greeting.java
```diff
@@ -0,0 +1,35 @@
...
+        } catch (InterruptedException e) {
+            System.err.println(e.getMessage());
+        }
+    }
+
+    /**
+     * This method wakes the thread.
+     */
+    void finishTheWelcoming() {
+        welcomer.interrupt();
```
TJB4epam, line 33: `Thread.currentThread.interrupt` should do from the inside of a thread.

src/test/resources/testng.xml
```diff
@@ -1,13 +1,26 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!DOCTYPE suite SYSTEM "https://testng.org/testng-1.0.dtd">
-
-<suite name="TestNG Group Test">
+<suite name="Tetris Suite" verbose="1">
```
TJB4epam, line 3: Why is "Tetris Test" just a "greeting test"?

Read on suite and test tags from testng.xml file - they carry meanings of set of tests and test scenario, respectively. The names confuse me here.

src/test/java/com/epam/prejap/tetris/greeting/GreetingTest.java
```diff
@@ -0,0 +1,36 @@
...
+
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Michał Cwynar
+ */
+@Test(groups = {"greeting"})
```
TJB4epam, line 11: All right. Time to get into nitty-gritty details of a multithreaded test realities.

This is no longer a deterministic code. Please have this done with special care.

TestNG has several features for multithreaded tests: thread pool, ability to run the code in a single thread or in multiple threads,  invocation count, multiple invocation of a threaded code methods etc. 

I'd like to see tests that use that here,to verify that multiple threads won't be an issue for the game.

TJB4epam: If you wish to keep the escaped reference (this in Greeting) then do test for whether that's an issue or not.

src/test/java/com/epam/prejap/tetris/greeting/GreetingTest.java
```diff
@@ -0,0 +1,36 @@
...
+
+    @DataProvider
+    public static Object[][] time() {
+        return new Object[][]{
+                {100},
+                {500},
+                {1000},
+                {2500},
+                {5000},
+                {10000}
```
TJB4epam, line 33: Again - numeric literals can use floor as a divider for better readability.

Additionally, I recommend reworking some of that code with `TimeUnit`.

TJB4epam (commented):
Requesting changes.

TJB4epam (changes requested):
Really nice job, and due to thread introduction I had to treat it especially harshly. So - number of changes to introduce, most worrying would be escaping reference and insufficient tests.

Additionally, older remarks are yet to be fully addressed.

TJB4epam:
Conflicts and older remarks. :(

TJB4epam:

pom.xml
Winetq: That's true, I don't need this dependency. Other versions moved to properties.

src/main/java/com/epam/prejap/tetris/Tetris.java
Winetq: Just because 10_000 :), but it can be changed.

src/main/java/com/epam/prejap/tetris/greeting/Greeting.java
Winetq: Javadocs was improved.

src/main/java/com/epam/prejap/tetris/greeting/Greeting.java
Winetq: That's a great idea, but maybe next time :).

src/main/java/com/epam/prejap/tetris/greeting/Welcomer.java
Winetq: It is needed because then the whole application will be killed.

src/main/java/com/epam/prejap/tetris/greeting/Welcomer.java
Winetq: We don't need this :)

src/main/java/com/epam/prejap/tetris/greeting/Welcomer.java
Winetq: Done.

src/main/java/com/epam/prejap/tetris/greeting/Greeting.java
Winetq: I have problems with the Thread.currentThread.interrupt(), because it works a little differently than the interrupting from another class like in the snippet above.

Winetq:
I couldn't refer to all remarks because of the lack of time, so we can close this PR.

marcin-ogorzalek-EPAM:
> I couldn't refer to all remarks because of the lack of time, so we can close this PR.

Closed as requested

PR #78, opened 2021-06-10
15 random obstacles on the field

ZytaWisz:
1. Added method that checks if command line arguments contain desired argument.
2. Added method that adds three random blocks to the bottom of the playfield.
3. Added test that checks if three random blocks were added to the bottom of the playfield.
4. Moved the plugins version to properties.


pom.xml
```diff
@@ -6,34 +6,45 @@
 
     <groupId>com.epam.prejap</groupId>
     <artifactId>tetris</artifactId>
-    <version>0.1</version>
+    <version>${version.tetris}</version>
```
TJB4epam, line 9: My memory may mislead me but I think this could be problematic to change. Have you tested this and made sure that Maven works with this addition?

pom.xml
```diff
@@ -6,34 +6,45 @@
...
 
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <maven.compiler.source>16</maven.compiler.source>
         <maven.compiler.target>16</maven.compiler.target>
+        <version.testng>7.4.0</version.testng>
```
TJB4epam, line 15: Kudos for that.

pom.xml
```diff
@@ -6,34 +6,45 @@
...
             <scope>test</scope>
         </dependency>
     </dependencies>
 
     <build>
         <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
```
TJB4epam, line 34: I think this disagrees with props version - there's 16 here we set 15. Am I correct in stating that's your IDE's doing?

pom.xml
```diff
@@ -6,34 +6,45 @@
...
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-jar-plugin</artifactId>
-                <version>3.2.0</version>
+                <version>${version.maven-jar-plugin}</version>
                 <configuration>
                     <archive>
                         <manifest>
                             <addClasspath>true</addClasspath>
-                            <mainClass>com.epam.prejap.git.tetris.Tetris</mainClass>
```
TJB4epam, line 36: I'd advise to drop that particular change as there's a PR already that does that along with tests. See #19

src/main/java/com/epam/prejap/tetris/Tetris.java
```diff
@@ -49,11 +51,25 @@ public static void main(String[] args) {
         var feed = new BlockFeed();
         var printer = new Printer(System.out);
         var playfield = new Playfield(rows, cols, feed, printer);
+
+        if (parseArgs(args, "addRandomBlocks")) {
```
TJB4epam, line 55: I'd like for entry argument to be documented. Where shall I look for the documentation for this one?

src/main/java/com/epam/prejap/tetris/game/Playfield.java
```diff
@@ -33,11 +35,11 @@ public void nextBlock() {
...
         boolean moved;
-            switch (move) {
-                case LEFT -> moveLeft();
-                case RIGHT -> moveRight();
-            }
-            moved = moveDown();
+        switch (move) {
+            case LEFT -> moveLeft();
+            case RIGHT -> moveRight();
+        }
```
TJB4epam, line 41: TY!

src/test/java/PlayfieldTest.java
```diff
@@ -0,0 +1,27 @@
...
+        //when
+        byte[][] grid = playfield.addRandomBlocks();
+
+        //than
+        Assert.assertTrue(Arrays.stream(grid)
+                .map(row -> IntStream.range(0, row.length).map(operand -> row[operand]).sum())
+                .mapToInt(Integer::intValue)
+                .sum() >= 12);
+    }
+}
```
TJB4epam, line 27: Heed GitHub's warning please.

src/test/java/PlayfieldTest.java
```diff
@@ -0,0 +1,27 @@
...
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+import java.util.Arrays;
+import java.util.stream.IntStream;
+
+public class PlayfieldTest {
+
+    @Test
+    public void testAddRandomBlocks() {
```
TJB4epam, line 13: Please, make the name a better match to what's tested here. I don't see us testing adding RANDOM blocks. :-)

src/main/java/com/epam/prejap/tetris/game/Playfield.java
```diff
@@ -105,6 +107,29 @@ private void forEachBrick(BrickAction action) {
         }
     }
 
+    /**
+     * Method adds three random blocks to the bottom of the playfield
+     *
+     * @author Zyta Wiszniewska
+     */
+    public byte[][] addRandomBlocks() {
```
TJB4epam, line 115: So, our PUBLIC API for this class now has `addRandomBlocks`... is this really what a play field should be doing? Asking for a friend.

TJB4epam (changes requested):
Several remarks need addressing.

pom.xml
ZytaWisz: You're right , it is problematic. I went back to the previous version

pom.xml
ZytaWisz: Yes, you're right. I cut that part of the code, and now it's okay

pom.xml
ZytaWisz: I went back to the previous version of code.

src/main/java/com/epam/prejap/tetris/Tetris.java
ZytaWisz: Done

src/main/java/com/epam/prejap/tetris/game/Playfield.java
ZytaWisz: No, I've change to private and invoked only during initialization of the Playfield

ZytaWisz:
I corrected the code. Pleas review @TJB4epam

src/main/java/com/epam/prejap/tetris/game/PlayFieldParameters.java
```diff
@@ -0,0 +1,12 @@
+package com.epam.prejap.tetris.game;
+
+import com.epam.prejap.tetris.block.BlockFeed;
+
+public record PlayFieldParameters(
```
TJB4epam, line 5: Very important thing: if something is public, it should be documented. And usually we don't want public things. Encapsulation suffers with public.

I like that it's a record. I am unsure if public is needed. Or: I believe without public it might be more useful.

Finally, I think that `boolean isWithRandomBlocks` means we have two VARIANTS of a play field. One with obstacles, one without. And this should be represented in types. Gotta go, so have to cut the review here.

TJB4epam (changes requested):
Think more of types when it comes to design.

Encapsulation. Responsibility. Why does PlayField exist? What problems it solves for us?

Class - Responsibility - Collaborators.

Also, interfaces are key.

src/main/java/com/epam/prejap/tetris/game/PlayFieldParameters.java
ZytaWisz: Public is needed as we need to create this record from Tetris class. 
What do you mean by 'And this should be represented in types' ?

ZytaWisz:
> Think more of types when it comes to design.
> 
> Encapsulation. Responsibility. Why does PlayField exist? What problems it solves for us?
> 
> Class - Responsibility - Collaborators.
> 
> Also, interfaces are key.

Could you please elaborate more on that ? I think that I don't get your point. If this is about the presence of 'addRandomBlocks' in PlayField class then I would say that this is the right place. Playfield creates (and it might be an empty or filled with random blocks field, doesn't matter really, it is a part of creation process) and manipulates the Play fileds of the game.

TJB4epam:
Conflicts reminder, please resolve.

src/main/java/com/epam/prejap/tetris/game/PlayFieldParameters.java
TJB4epam: Answered in Teams, but in short: we have a variant without random blocks and with. And they behave differently. So, perhaps we could model this using types? Inheritance? Composition? Some pattern perhaps?

TJB4epam (changes requested):
The largest gripe I have with these changes is that they make Playfield do too much. 

IMO randomizing obstacles should happen OUTSIDE of Playfield and since obstacles are blocks, we should be using BlockFeed. 

So, I'd see this change as done OUTSIDE of Playfield and PASSED IN - *IF REQUIRED*.

If we want to recreate this as a playfield with obstacles or playfield without then I'd insist on using types to have two different classes. 

But again, I'd see this as done OUTSIDE and PASSED IN.

PR #81, opened 2021-06-11
10-current-score-implementation

PatSew:
Hello,

Seems to be working but I am not really satisfied with my implementation. It is very ugly and simple.

Is there anything I could test in such case?

Any hints appreciated!


pom.xml
```diff
@@ -38,6 +38,11 @@
                     </archive>
                 </configuration>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <version>3.8.1</version>
```
TJB4epam, line 44: Please put that in properties.

Also, bump the version in the POM, otherwise this can't go it.

TJB4epam (commented):
POM - version bump is missing while required.
No tests, can't go in.
Score should be an object, not be a couple of printlns here or there. That way we will be able to test it and printing it will be just a call to it's `toString`.

TJB4epam (approved):
POM - version bump is missing while required.
No tests, can't go in.
Score should be an object, not be a couple of printlns here or there. That way we will be able to test it and printing it will be just a call to it's `toString`.

Previous comment indicated approval, but it's a request for changes.

Also, please resolve conflicts.

TJB4epam (changes requested):
POM - version bump is missing while required.
No tests, can't go in.
Score should be an object, not be a couple of printlns here or there. That way we will be able to test it and printing it will be just a call to it's `toString`.

Previous comment indicated approval, but it's a request for changes.

Also, please resolve conflicts.

Issue #70, opened 2021-06-07
Make blocks have different colors


AndriiVel:
As a user I want an elegant and nice-looking program.
Make every new block to have different colors.

Example:
```
$ mvn clean package
[...]
$ java -jar target/tetris-0.1.jar
```
![image](https://user-images.githubusercontent.com/63800077/120998925-d3edf880-c788-11eb-9e6e-624847bc564d.png)

KoLaR8:
I'm working on this task.

PR #82, opened 2021-06-12
70 implement blocks with different colors

KoLaR8:
Solve #70 issue. Each block type has a different color.
obraz

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,8 @@
     <modelVersion>4.0.0</modelVersion>
     <groupId>com.epam.prejap</groupId>
     <artifactId>tetris</artifactId>
-    <version>0.16</version>
+    <version>0.17</version>
+
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <maven.compiler.source>16</maven.compiler.source>
--- a/src/main/java/com/epam/prejap/tetris/block/Block.java
+++ b/src/main/java/com/epam/prejap/tetris/block/Block.java
@@ -1,6 +1,7 @@
 package com.epam.prejap.tetris.block;
 
 import com.epam.prejap.tetris.logger.Logger;
+import com.epam.prejap.tetris.game.Color;
 
 public abstract class Block {
     private static final Logger LOGGER = Logger.getLogger(Block.class);
@@ -8,6 +9,7 @@ public abstract class Block {
     private final byte[][] image;
     private final int rows;
     private final int cols;
+    Color color = Color.WHITE;
 
     Block(byte[][] dots) {
         rows = dots.length;
@@ -28,7 +30,10 @@ public abstract class Block {
                     LOGGER.error("The dot value for block is less than 0. The block cannot be created");
                     throw new IllegalArgumentException("Invalid dot value");
                 }
-                image[i][j] = dot;
+                if(dot > 0){
+                    color = Color.getColorById(dot);
+                    image[i][j] = color.id;
+                }
             }
         }
     }
@@ -45,4 +50,7 @@ public abstract class Block {
         return image[i][j];
     }
 
+    public Color color() {
+        return color;
+    }
 }
--- a/src/main/java/com/epam/prejap/tetris/block/TBlock.java
+++ b/src/main/java/com/epam/prejap/tetris/block/TBlock.java
@@ -1,18 +1,19 @@
 package com.epam.prejap.tetris.block;
 
 import com.epam.prejap.tetris.logger.Logger;
+import com.epam.prejap.tetris.game.Color;
 
 /**
  * Generates a "T" block in the Tetris game - when "1" in a byte array is present then a "#" is printed,
  * otherwise a whitespace (" ") is printed.
- * @see com.epam.prejap.tetris.game.Printer#print(byte)
+ * @see com.epam.prejap.tetris.game.Printer#print(int)
  * @author Grzegorz Konopka
  */
 final class TBlock extends Block {
     private static final Logger LOGGER = Logger.getLogger(TBlock.class);
-
+    private static final byte COLOR_ID = Color.GREEN.id;
     /**
-     * Byte array represents "Y" block.
+     * Byte array represents "T" block.
      * In game will be displayed as (between lines):
      *  -------------------
      *      ###
@@ -20,8 +21,8 @@ final class TBlock extends Block {
      *  -------------------
      */
     private static final byte[][] IMAGE = {
-            {1,1,1},
-            {0,1,0}
+            {COLOR_ID, COLOR_ID, COLOR_ID},
+            {0, COLOR_ID, 0}
     };
 
     TBlock() {
@@ -29,4 +30,3 @@ final class TBlock extends Block {
         LOGGER.trace("{} was created", getClass().getSimpleName());
     }
 }
-
--- a/src/main/java/com/epam/prejap/tetris/block/HBlock.java
+++ b/src/main/java/com/epam/prejap/tetris/block/HBlock.java
@@ -1,17 +1,19 @@
 package com.epam.prejap.tetris.block;
 
 import com.epam.prejap.tetris.logger.Logger;
+import com.epam.prejap.tetris.game.Color;
 
 /**
  * Implements H shaped block
- * @see com.epam.prejap.tetris.game.Printer#print(byte) 
  * @author Slawomir Kucab
+ * @see com.epam.prejap.tetris.game.Printer#print(int)
  */
 final class HBlock extends Block{
     private static final Logger LOGGER = Logger.getLogger(HBlock.class);
+    private static final byte COLOR_ID = Color.CYAN.id;
 
     /**
-     * Byte array represents "S" block.
+     * Byte array represents "H" block.
      * In game will be displayed as (between lines):
      *  -------------------
      *       # #
@@ -20,13 +22,12 @@ final class HBlock extends Block{
      *  -------------------
      */
     private static final byte[][] IMAGE = {
-            {1, 0, 1},
-            {1, 1, 1},
-            {1, 0, 1}
+            {COLOR_ID, 0, COLOR_ID},
+            {COLOR_ID, COLOR_ID, COLOR_ID},
+            {COLOR_ID, 0, COLOR_ID}
     };
     public HBlock(){
         super(IMAGE);
         LOGGER.trace("{} was created", getClass().getSimpleName());
     }
 }
-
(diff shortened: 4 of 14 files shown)


src/main/java/com/epam/prejap/tetris/block/Block.java
```diff
@@ -1,11 +1,13 @@
...
 
+import com.epam.prejap.tetris.colours.ColorPicker;
+
 public abstract class Block {
 
     final byte[][] image;
     final int rows;
     final int cols;
-
+    final String color = ColorPicker.chooseRandomColor();
```
TJB4epam, line 10: Is that it? Random colour each time? Was that mentioned in story card? My memory may mislead me but I thought it's colour assigned to a block?

src/main/java/com/epam/prejap/tetris/block/BlockFeed.java
```diff
@@ -8,7 +8,8 @@
 
     private final Random rnd = new Random();
     private final List<Supplier<Block>> blocks = List.of(
-        OBlock::new
+            OBlock::new,
```
TJB4epam, line 11: Not that I particularly mind, but why that additional indent?

src/main/java/com/epam/prejap/tetris/block/TBlock.java
```diff
@@ -0,0 +1,20 @@
+package com.epam.prejap.tetris.block;
```
TJB4epam, line 1: Why is this reported as a new file that you're adding if it was added and merged by Grzegorz earlier?

src/main/java/com/epam/prejap/tetris/colours/ColorPicker.java
```diff
@@ -0,0 +1,63 @@
+package com.epam.prejap.tetris.colours;
+
+import java.util.Random;
+
+/**
+ *  Purpose of this class is to menage different colors  behaviour provided from enum Colors.
```
TJB4epam, line 6: Menage - French.
Manage - English. :-)

Also, there are two spaces between colours and behaviour.

TJB4epam: Also - that's a rather vague name. I still don't know what this does, aside from "deals with colours", which I have from class name. Any better description?

src/main/java/com/epam/prejap/tetris/colours/ColorPicker.java
```diff
@@ -0,0 +1,63 @@
...
+ * @author Wojciech Kolarczyk
+ * @see Colors
+ */
+public class ColorPicker {
```
TJB4epam, line 10: A curious case, here.

This class has NO data and all of it's API is `public static`. So, this is a tool class. However it works on an enum and accepts data from outside. 

Why then don't we move those methods to `enum Colours`?

If not, at least let's discourage instantiation of this class by adding a private default c-tor, with a comment, as per Effective Java.

src/main/java/com/epam/prejap/tetris/colours/Colors.java
```diff
@@ -0,0 +1,44 @@
+package com.epam.prejap.tetris.colours;
```
TJB4epam, line 1: Not sure a part of UI of our game should be hidden in a package. Frankly, I fail to see the benefits this gives us.

I'd rather have this mixed with the UI package, or mixed into a package our UI hides / is used in.

Also, I have a feeling that if we combined this and the static method class ColorPicker we could deal away with 'getID' and 'getAsciCode' or however they are called.

Finally: ASCII, so asciiSomething, not asciSomething. Double i, even when in var name.

TJB4epam (changes requested):
Two large issues in a lovely approach to a nice issue.

1. Number of files I read here are seemingly new, while they aren't. This makes me confused about whether I'm reviewing YOUR code or whether I'm re-reviewing somebody else's code. Please straighten this out. `git rebase` should help.
2. I have a design doubt w.r.t. colours package. See my review remarks for details.

Thank you, requesting changes.

src/main/java/com/epam/prejap/tetris/block/Block.java
KoLaR8: I was solving issue #70 , which says about new color for each block. After I actually write this code I saw that issue #12 is also about coloring blocks in game.

KoLaR8:
Changes made:
1. moved content of ColorPicker class to enum Colors. 
2. Deleted ColorPicker class. 
3. Moved Colors to game package. 
4. Correct typos with 'asci' to "ascii", and with "menage" to "manage". 
5. Removed additional indent in BlockFeed.
6. Changed tests for OBlock

pom.xml
```diff
@@ -6,7 +6,7 @@
 
     <groupId>com.epam.prejap</groupId>
     <artifactId>tetris</artifactId>
-    <version>0.4</version>
+    <version>0.5</version>
```
TJB4epam, line 9: TY!

src/main/java/com/epam/prejap/tetris/block/Block.java
```diff
@@ -39,4 +42,8 @@ public byte dotAt(int i, int j) {
         return image[i][j];
     }
 
+    public String  color(){
```
TJB4epam, line 45: Colour is a String? :-O

src/main/java/com/epam/prejap/tetris/block/BlockFeed.java
```diff
@@ -8,8 +8,8 @@
 
     private final Random rnd = new Random();
     private final List<Supplier<Block>> blocks = List.of(
-            OBlock::new,
-            TBlock::new
+        OBlock::new,
+        TBlock::new
```
TJB4epam, line 12: TY

src/main/java/com/epam/prejap/tetris/block/Block.java
TJB4epam: OK, ASCII codes, OK.

src/main/java/com/epam/prejap/tetris/game/Colors.java
```diff
@@ -0,0 +1,81 @@
+package com.epam.prejap.tetris.game;
+
+import java.util.Random;
+
+/**
+ * Contains definition of colors used in tetris game to print blocks in different colors.
+ * Each value contains ascii code and id.
+ * @author Wojciech Kolarczyk
+ */
+public enum Colors {
```
TJB4epam, line 10: Plus for using ASCII colours, I don't think I mentioned that before but that's a nice move.

src/test/java/com/epam/prejap/tetris/block/OBlockTest.java
```diff
@@ -1,5 +1,6 @@
 package com.epam.prejap.tetris.block;
```
TJB4epam, line 1: Once I merge Anna's changes these tests will have to be redone.

src/main/java/com/epam/prejap/tetris/block/Block.java
```diff
@@ -1,11 +1,13 @@
...
 
+import com.epam.prejap.tetris.game.Colors;
+
 public abstract class Block {
 
     final byte[][] image;
     final int rows;
     final int cols;
-
+    final String color = Colors.chooseRandomColor();
```
TJB4epam, line 10: What I dislike about this is how we're using partial data out of the object, instead of relying on an object.

Why not have a Colours enum object here instead of a String?

src/main/java/com/epam/prejap/tetris/game/Colors.java
```diff
@@ -0,0 +1,81 @@
...
+    public static String printHash(byte id){
+        String color = Colors.WHITE.asciiCode;
+        for(var colors: Colors.values()){
+            if(colors.id == id){
+                color = colors.asciiCode;
+            }
+        }
+        return color + "#" + Colors.RESET.asciiCode;
+    }
+}
```
TJB4epam, line 81: If this enum had a decent toString method, we could work on it, like so:

WHITE.toString()

RED.toString()

RESET.toString()

in this manner, plenty of code that relies on a string from that class, could use type-safety, by using DIRECTLY some constant from this class. Like, WHITE or GREEN. Additional bonus comes in increased readability.

src/main/java/com/epam/prejap/tetris/game/Colors.java
```diff
@@ -0,0 +1,81 @@
...
+     * @param asciiCode needed to change color of printed message in console.
+     * @param id needed to recognize color of block in playfield.
+     */
+    Colors(String asciiCode, byte id){
+        this.asciiCode = asciiCode;
+        this.id = id;
+    }
+
+
+    public static String chooseRandomColor(){
```
TJB4epam, line 34: Try moving away from statics in favour of methods on an object. Enum is one, so let's use it. You have the object with a state (ASCII code, id) and you can use this in methods - you already are.

src/main/java/com/epam/prejap/tetris/game/Colors.java
```diff
@@ -0,0 +1,81 @@
...
+     * Used to change '1' to block's color id in block image.
+     * @param color name of color.
+     * @return  id of color.
+     */
+    public static byte id(String color) {
+        for(var colors : Colors.values()){
+            if(colors.asciiCode.equals(color)){
+                return colors.id;
+            }
+        }
```
TJB4epam, line 63: Much better complexity here can be achieved with EnumMap. Instead of O(n) where n is number of colours we would have O(1).

TJB4epam (changes requested):
Take a look at the comments. I like the direction and this is better than previous version by far. We can lean on enums more and use them in a better, type-safer way.

KoLaR8:
Changes made:
1. Rename Colors enum to Color.
2. Created toString method for Color.
3. Introduced Map for faster searching through Color values.
4. Replaced all string which were ascii codes with Color value.
5. Added messages to assertions in tests.
6. Changed tests in BlockShapesTest to fit my implementation. (Checking for color id, not for '1').
7. Added test for map.

TJB4epam:

KoLaR8:
@TJB4epam 
updated to latest version with pom version to 11. Latest changes made me force to change tactic from generating random colour to block shape's defined colour, but it's much better and more tetris alike now. Please review as soon as possible, as this change is little bigger than others.

TJB4epam (changes requested):
See comments. Especially the one asking for complexity reduction - the longest.

AndriiVel:
Approved during the lecture. Version: 0.13

AndriiVel:
A follow-up issue will be created later (Complexity of feature, toStrings, enumMap)

marcin-ogorzalek-EPAM:
@KoLaR8 merge conflicts to solve

pkierat-epam:
Looks good. Merging.

PR #85, opened 2021-06-16
Elapsed time is present #16

Dawid126:
Implemented Timer class for measuring elapsed time from the beginning of the game.
Tests are included and all pass.
I have already implemented the functionality of pausing and resuming the timer, because there is an issue present about implementing the posibility to pause the game.


Dawid126:
@TJB4epam ready to review

pom.xml
```diff
@@ -6,7 +6,7 @@
 
     <groupId>com.epam.prejap</groupId>
     <artifactId>tetris</artifactId>
-    <version>0.4</version>
+    <version>0.5</version>
```
TJB4epam, line 9: Thank you, but I have a feeling this should be higher? Please check.

src/main/java/com/epam/prejap/tetris/game/Timer.java
```diff
@@ -0,0 +1,69 @@
+package com.epam.prejap.tetris.game;
```
TJB4epam, line 1: I am unsure of this placement. As in, I don't see the benefits. Clearly we call it only from `Tetris` class, which is in a different package. So... why here?

src/main/java/com/epam/prejap/tetris/game/Timer.java
```diff
@@ -0,0 +1,69 @@
+package com.epam.prejap.tetris.game;
+
+import java.text.SimpleDateFormat;
+import java.time.Duration;
+import java.time.LocalTime;
+
+/**
+ * Allows to track time from a given point on timeline.
```
TJB4epam, line 8: Nice JD and I like the pause/resume options. TY!

TJB4epam: Minor English issue: on a timeline, article is missing.

src/main/java/com/epam/prejap/tetris/game/Timer.java
```diff
@@ -0,0 +1,69 @@
...
+import java.time.LocalTime;
+
+/**
+ * Allows to track time from a given point on timeline.
+ * The time format is "HH:mm:ss".
+ * There is also an option to pause the timer and resume it.
+ *
+ * @author Dawid Białka
+ */
+public class Timer {
```
TJB4epam, line 14: Maybe it doesn't need to be `public`, see comment I left on a package declaration here.

src/test/java/com/epam/prejap/tetris/timer/TimerTest.java
```diff
@@ -0,0 +1,85 @@
...
+import org.testng.annotations.Test;
+
+import java.time.LocalTime;
+
+import static org.testng.Assert.*;
+
+/**
+ * @author Dawid Białka
+ */
+
```
TJB4epam, line 15: Please cut.

src/test/java/com/epam/prejap/tetris/timer/TimerTest.java
```diff
@@ -0,0 +1,85 @@
...
+@Test(groups = {"timer"})
```
TJB4epam, line 16: Several remarks to that class:

1. All tests with thread sleep should have a timeout, so that accidental mistake with making the thread sleep for an hour would NOT make all tests run for an hour. :p 
2. I'd like to create a new group - timeoutPossible - and add all tests with thread sleep to it. The tests with timeouts are special.
3. Move all tests which expect to throw RTE to a separate class. There's something fishy about throwing, but we don't yet have logging, so I'm keeping my remarks FOR NOW. You may want to collaborate with @AndriiVel who's coding a logger.
4. Mark these tests as integration tests - you are checking the time flow, that's not just "unit". There's collaboration when threading or timing as both require operating system.

TJB4epam (changes requested):
See my comments. Several issues to take care of.

TJB4epam:
Conflicts in the testng.xml file.

src/test/java/com/epam/prejap/tetris/timer/TimerTest.java
Dawid126: So should I continue as if there were no logger available or should I pull/rebase to Andrii branch and use the logger in my code even though Andrii's work hasn't been yet merged to master?

TJB4epam: Options are as you listed, plus continue this as it is AND fix this later once logger is "in".

Re exceptions: throwing crashes the program. That's an option, but that's fail-fast. There's also fail-safe, where we do NOT crash but do not start the timer / pause the timer when it's not allowed. Then, logging a warning would suffice while ignoring the method call.

TJB4epam:
Conflicts, old remarks to address.

src/main/java/com/epam/prejap/tetris/timer/Timer.java
```diff
@@ -0,0 +1,70 @@
...
+import java.text.SimpleDateFormat;
+import java.time.Duration;
+import java.time.LocalTime;
+
+/**
+ * Allows to track time from a given point on a timeline.
+ * The time format is "HH:mm:ss".
+ * There is also an option to pause the timer and resume it.
+ */
+
```
marcin-ogorzalek-EPAM, line 12: No author and unnecessary empty line between class and jd

src/main/java/com/epam/prejap/tetris/timer/Timer.java
```diff
@@ -0,0 +1,70 @@
+package com.epam.prejap.tetris.timer;
```
marcin-ogorzalek-EPAM, line 1: Why not simply in `com.epam.prejap.tetris`? What is the benefit of the dedicated package with just one class inside?

src/main/java/com/epam/prejap/tetris/timer/Timer.java
```diff
@@ -0,0 +1,70 @@
...
+import java.time.Duration;
+import java.time.LocalTime;
+
+/**
+ * Allows to track time from a given point on a timeline.
+ * The time format is "HH:mm:ss".
+ * There is also an option to pause the timer and resume it.
+ */
+
+public class Timer {
```
marcin-ogorzalek-EPAM, line 13: `public` connected with https://github.com/EPAM-Online-Courses/tetris/pull/85/files#r658743717

src/test/java/com/epam/prejap/tetris/timer/TimerIT.java
```diff
@@ -0,0 +1,41 @@
...
+import org.testng.annotations.Test;
+
+import java.time.LocalTime;
+
+import static org.testng.Assert.assertEquals;
+
+/**
+ * Integration tests that check whether the timer is measuring the time correctly.
+ */
+
```
marcin-ogorzalek-EPAM, line 14: Definitely unnecessary empty line

src/test/java/com/epam/prejap/tetris/timer/TimerIT.java
```diff
@@ -0,0 +1,41 @@
...
+import org.testng.annotations.BeforeMethod;
+import org.testng.annotations.Test;
+
+import java.time.LocalTime;
+
+import static org.testng.Assert.assertEquals;
+
+/**
+ * Integration tests that check whether the timer is measuring the time correctly.
+ */
```
marcin-ogorzalek-EPAM, line 13: This javadoc has no author. The more important thing is that we should avoid creating those for tests. This piece of documentatiotion doesn't provide anything that we cannot read from the tests.

src/test/java/com/epam/prejap/tetris/timer/TimerIT.java
```diff
@@ -0,0 +1,41 @@
...
+
+@Test(groups = "timeoutPossible")
+public class TimerIT {
+
+    private Timer timer;
+
```
marcin-ogorzalek-EPAM, line 19: One empty line too much.

src/test/java/com/epam/prejap/tetris/timer/TimerIT.java
```diff
@@ -0,0 +1,41 @@
...
+import java.time.LocalTime;
+
+import static org.testng.Assert.assertEquals;
+
+/**
+ * Integration tests that check whether the timer is measuring the time correctly.
+ */
+
+@Test(groups = "timeoutPossible")
+public class TimerIT {
```
marcin-ogorzalek-EPAM, line 16: Are those really the integration tests? We have only one class in use in the thread. No connection to outside classes in the app.

src/main/java/com/epam/prejap/tetris/timer/Timer.java
```diff
@@ -0,0 +1,70 @@
...
+        Duration timeElapsed = Duration.between(startTime, currentTime);
+
+        return LocalTime.of(
+                (int) timeElapsed.toHours(),
+                timeElapsed.toMinutesPart(),
+                timeElapsed.toSecondsPart());
+    }
+
+    public void showElapsedTime() {
+        System.out.println("Elapsed time: " + this.getElapsedTime());
```
marcin-ogorzalek-EPAM, line 67: Instead of sout wit concatenation inside please use souf.

src/test/java/com/epam/prejap/tetris/timer/TimerUT.java
```diff
@@ -0,0 +1,99 @@
...
+
+    @BeforeTest
+    void setLineSeparator() {
+        String os = System.getProperty("os.name");
+        if(os.contains("Windows")) {
+            lineSeparator = "\r\n";
+        } else {
+            lineSeparator = "\n";
+        }
+    }
```
marcin-ogorzalek-EPAM, line 27: There is something like `System.lineSeparator()` in Java that does exactly this thing here.

src/test/java/com/epam/prejap/tetris/timer/TimerUT.java
```diff
@@ -0,0 +1,99 @@
...
+        timer.getElapsedTime();
+
+        String expectedOutput = "Warning: Timer is paused" + lineSeparator;
+        assertEquals(consoleContent.toString(), expectedOutput);
+    }
+
+
+    @AfterTest
+    public void afterTest() {
+        System.setErr(this.originalStdErr);
```
marcin-ogorzalek-EPAM, line 90: Why this is cleared after all the tests, while it is set before each test?

src/test/java/com/epam/prejap/tetris/timer/TimerUT.java
```diff
@@ -0,0 +1,99 @@
...
+    public void isStartedShouldReturnTrue() {
+        timer.start();
+        assertTrue(timer.isStarted());
+    }
+
+    public void startShouldOPrintWarning() {
+        timer.start();
+        timer.start();
+
+        String expectedOutput = "Warning: Timer is already started" + lineSeparator;
```
marcin-ogorzalek-EPAM, line 48: `String.format` IMO - line separator is still a string.

src/test/java/com/epam/prejap/tetris/timer/TimerIT.java
marcin-ogorzalek-EPAM: Nevermind as per Tomasz's comment

marcin-ogorzalek-EPAM:
Bump up pom version

src/test/java/com/epam/prejap/tetris/timer/TimerUT.java
```diff
@@ -0,0 +1,99 @@
...
+
+import org.testng.annotations.*;
+
+import java.io.ByteArrayOutputStream;
+import java.io.PrintStream;
+
+import static org.testng.Assert.*;
+
+@Test(groups = "timerUT")
+public class TimerUT {
```
marcin-ogorzalek-EPAM, line 11: This can be simpy named as `TimerTest` as those are unit tests.

src/main/java/com/epam/prejap/tetris/timer/Timer.java
Dawid126: But how to test it when I make the class package private and the test class is in another package?

src/test/java/com/epam/prejap/tetris/timer/TimerUT.java
Dawid126: Changed it and now I use the reset method on ByteArrayOutputStream after each test and set the Err once before all the tests.

src/test/java/com/epam/prejap/tetris/timer/TimerIT.java
pkierat-epam: @marcin-ogorzalek-EPAM What do you mean by that?

src/main/java/com/epam/prejap/tetris/Timer.java
```diff
@@ -0,0 +1,79 @@
...
+public class Timer {
+    private static final Logger LOGGER = Logger.getLogger(Timer.class);
+
+    private LocalTime startTime;
+    private LocalTime stopTime;
+    private boolean isPaused = false;
+
+    public void start() {
+        if (this.startTime != null) {
+            System.err.println("Warning: Timer is already started");
```
pkierat-epam, line 24: The game has a dedicated component serving as a UI, called `Printer`. Yet your class writes directly to standard error stream, making testing a lot harder, since you need to change a global variable `System.err`.

src/main/java/com/epam/prejap/tetris/Timer.java
```diff
@@ -0,0 +1,79 @@
...
+        return LocalTime.of(
+                (int) timeElapsed.toHours(),
+                timeElapsed.toMinutesPart(),
+                timeElapsed.toSecondsPart());
+    }
+
+    public void showElapsedTime() {
+        LocalTime elapsedTime = this.getElapsedTime();
+        LOGGER.trace("Printed elapsed time: ", elapsedTime);
+        System.out.format("Elapsed time: %s", elapsedTime).println();
```
pkierat-epam, line 77: ... and here you do an even worse thing - print the elapsed time directly to stdout, completely disregarging the `Printer`.

src/test/java/com/epam/prejap/tetris/timer/TimerTest.java
```diff
@@ -0,0 +1,94 @@
...
+
+    private Timer timer;
+    private final PrintStream originalStdErr = System.err;
+    private final ByteArrayOutputStream consoleContent = new ByteArrayOutputStream();
+    private String lineSeparator;
+
+    @BeforeTest
+    void setLineSeparatorAndStdErr() {
+        lineSeparator = System.lineSeparator();
+        System.setErr(new PrintStream(this.consoleContent));
```
pkierat-epam, line 22: Capturing `System.err` makes running the tests in parallel unreliable (related to the previous remarks). This should be solved rather by introducing `Printer` as a dependency for `Timer` and mocking it in the tests.

pkierat-epam (changes requested):
To be changed:
* Use Printer to interact with a user instead of writing directly to stdout/stderr.

PR #87, opened 2021-06-17
Implementation for Issue #8

michal-majewski:
Add printing of the next block's Hint.

I am asking for a review, especially about the way of solution. I was trying to resolve this using interfaces and extended class, but finally, I have decided to go this way.


src/main/java/com/epam/prejap/tetris/game/Playfield.java
```diff
@@ -12,6 +12,7 @@
     private final BlockFeed feed;
 
     private Block block;
+    private Block nextBlock;
```
TJB4epam, line 15: Don't forget to add the POM file with an increased version. That's a conflict resolution exercise after all! :-)

Without it we won't let it in. :)

src/main/java/com/epam/prejap/tetris/game/Playfield.java
```diff
@@ -24,7 +25,12 @@ public Playfield(int rows, int cols, BlockFeed feed, Printer printer) {
     }
 
     public void nextBlock() {
-        block = feed.nextBlock();
+        if (nextBlock == null) {
```
TJB4epam, line 28: How about a ternary op instead?

`block = (nextBlock == null) ? feed.nextBlock() : nextBlock`

src/main/java/com/epam/prejap/tetris/game/Playfield.java
TJB4epam: Also, this time commenting on the actual line the comment appears on:

`nextBlock` is a name already used in that class. Consider a better one, for `nextBlock` already carries meaning here in Playfield`.

TJB4epam: Issue lingo suggests:

1. hintBlock
2. newBlock

I prefer the first one.

src/main/java/com/epam/prejap/tetris/game/Playfield.java
```diff
@@ -24,7 +25,12 @@ public Playfield(int rows, int cols, BlockFeed feed, Printer printer) {
     }
 
     public void nextBlock() {
```
TJB4epam, line 27: With this approach we're making it the issue of a PLAYFIELD to obtain the information about NEXT BLOCK AND HINTED BLOCK.

Why it's not a BlockFeed's specialty? After all, the source of the blocks is there, not here, in Playfield.

Options I see:

1. BlockFeed returns a list of blocks, next one and hinted one.
2. BlockFeed has an API to hint, that actually plays well with randomizing a new block (as in, if hint says next block is a T-block then it IS a T-block and we won't suddenly re-randomize it).
3. We keep the responsibility for hinting inside Playfield, making it one step close to anti-pattern Sun God Class. :P (If the sarcasm isn't clear, that option is least recommended).
4. There are a few more options we could use, like introducing a HintedBlock type, which BlockFeed would randomize, then we could use nextBlock method here or on HintedBlock to convert the HintedBlock type into a Block type.

src/main/java/com/epam/prejap/tetris/game/Printer.java
```diff
@@ -1,17 +1,36 @@
 package com.epam.prejap.tetris.game;
 
+import com.epam.prejap.tetris.block.Block;
+
 import java.io.PrintStream;
 
 public class Printer {
 
-    final PrintStream out;
+    private static final String NEXT_BLOCK_HINT_TEXT = "\u2610 NEXT BLOCK! \u2610";
```
TJB4epam, line 9: Consider talking to @KOlaR8 - Wojtek - about his colours impl. I see you guys may want to collab a bit with regard to that part. :-)

src/test/java/com/epam/prejap/tetris/game/PrinterTest.java
```diff
@@ -0,0 +1,91 @@
...
+                    stringBlock.append(" ");
+                } else {
+                    stringBlock.append("#");
+                }
+            }
+            stringBlock.append("\n");
+        }
+        return stringBlock.toString();
+    }
+}
```
TJB4epam, line 91: Line ending / file ending.

src/test/java/com/epam/prejap/tetris/game/PrinterTest.java
```diff
@@ -0,0 +1,91 @@
...
+        expectedOutput = null;
+        printer = null;
+        hintBlock = null;
+        outContent = null;
+    }
+
+    public void shouldPrintProperHint() {
+        //given
+        expectedOutput = """
+                +--------------------+
```
TJB4epam, line 44: Multi-line text used rightly. See Anna's changes on block tests though.

src/test/java/com/epam/prejap/tetris/game/PrinterTest.java
```diff
@@ -0,0 +1,91 @@
...
+import org.testng.annotations.AfterMethod;
+import org.testng.annotations.BeforeMethod;
+import org.testng.annotations.Test;
+
+import java.io.ByteArrayOutputStream;
+import java.io.PrintStream;
+
+/**
+ * @author Michal Majewski
+ * @version 1.0
```
TJB4epam, line 15: If you're using version in JD make it consistent with version in the POM please. Also, why in tests?

src/main/java/com/epam/prejap/tetris/game/Printer.java
```diff
@@ -23,23 +42,56 @@ void draw(byte[][] grid) {
...
+        out.print(NEXT_BLOCK_HINT_TEXT);
+        out.print(" ".repeat(2));
+        endRow();
+        border(20);
+    }
+
+    private void printBlockShape(final Block hintBlock) {
+        var stringBlock = new StringBuilder();
+
+        for (var i = 0; i < hintBlock.rows(); i++) {
```
TJB4epam, line 84: Is this the responsibility of a printer or a good toString on a block? Asking for a friend.

src/test/java/com/epam/prejap/tetris/game/PrinterTest.java
```diff
@@ -0,0 +1,91 @@
...
+ **/
+@Test(groups = "HintPrinter")
```
TJB4epam, line 17: FYI, that's an integration tests. Printer, Java API, screen - how they integrate is what we test. Please mark it as such and add it to a necessary group.

TJB4epam: I like the fact grouping of tests happens at the top. Quick glance and I'll know the group here. A hint: I'd actually make this a part of printing group, if there's such a group. You know, for testing all the things print-related.

src/test/java/com/epam/prejap/tetris/game/PrinterTest.java
```diff
@@ -0,0 +1,91 @@
...
+        printer.printTheNextBlockHint(null);
+
+        //then
+        Assert.assertEquals(outContent.toString(), expectedOutput);
+
+    }
+
+    //FIXME: find another solution to make the checking available with random blocks.
+    // Now blocks are available only via the BlockFeed class.
+    private String blockToString(Block block) {
```
TJB4epam, line 75: To test random, you need your own randomizer, that's cooperating enough. The built-in in Java isn't - too pseudorandom. ;)

So, I'd suggest your own class that extends it, to be used for testing purposes.

TJB4epam (changes requested):
I like how it looks, you know. The hint header is clear, distinctive, close enough to play field area and far enough to NOT be mistaken for it. Perhaps one more blank line between them and we're cool.

Code has several points of note, so do tests. See remarks. Requesting changes.

TJB4epam:
Still no replies to my remarks, still no POM version update, and new conflicts. :(

michal-majewski:
I'll do it this week. I was struggling with job duties. I promise

src/main/java/com/epam/prejap/tetris/game/Playfield.java
michal-majewski: I'll choose the second option. The BlockFeed as a feeder of blocks should provide all information about blocks. The Playfield, as you said is not responsible for checking which block is the current block and which is the next one. I'll move this part of functionality to the BlockFeed class.

src/main/java/com/epam/prejap/tetris/game/Playfield.java
michal-majewski: Sure, It will be much better to call this variable a "hintBlock". More readable.

src/test/java/com/epam/prejap/tetris/game/PrinterTest.java
michal-majewski: Ok, I'll make JD in classes consistent with the POM. Unfortunately, in tests, it was added automatically, and I did not take care of it. I'll remove it in tests.

src/main/java/com/epam/prejap/tetris/game/Printer.java
michal-majewski: It is a nice idea and a nice feature. I will gladly cooperate with Wojtek. I hope it can be done in the new PR.

michal-majewski:
Hello, please let me know if provided changes are sufficient. If not, I'm asking for remarks. I'll try to fix them after work.

pkierat-epam:
There are conflicts to be resolved.

michal-majewski:
Yes, I know. I am working on it. It is almost done.

michal-majewski:
Ok, I have fixed the occurring warnings and errors during the packaging

PR #92, opened 2021-06-18
10-present-score-at-every-block-screen

rpiw:
Add one method to capture current score and print it with additional message. I do not think any tests are necessary here, because it 3 lines of coding: 1 line for capturing score, 1 for creating a fancy formatted string (I could moved it into sout, but it would be even too short for a feature), 1 for printing. For issue, see #10


src/main/java/com/epam/prejap/tetris/Tetris.java
```diff
@@ -28,19 +28,22 @@ public Score play() {
 
             playfield.nextBlock();
             score++;
-
             boolean nextMove;
             do {
                 waiter.waitForIt();
                 Move move = player.nextMove().orElse(Move.NONE);
                 moved |= (nextMove = playfield.move(move));
+                scoreListener(new Score(score));
```
marcin-ogorzalek-EPAM, line 36: Well, there is not much going here and this cannot be tested in a unit test. Integrational one is the other thing. There is a possibility to capture the output to check if it is correct. `ByteArrayOutputStream` and `System.setOut` can be helpful.

marcin-ogorzalek-EPAM: But this may be not necessary. That depends on what @TJB4epam will say

src/main/java/com/epam/prejap/tetris/Tetris.java
```diff
@@ -28,19 +28,22 @@ public Score play() {
...
             } while (nextMove);
-
         } while (moved);
 
         return new Score(score);
     }
 
+    private static void scoreListener(Score score) {
```
marcin-ogorzalek-EPAM, line 43: To be honest. Is the main focus of this method is to listen for sore or prints is to the users in a fancy way?

src/main/java/com/epam/prejap/tetris/Tetris.java
```diff
@@ -28,19 +28,22 @@ public Score play() {
...
+        System.out.printf("Your score is %s%n", score.points());
```
marcin-ogorzalek-EPAM, line 44: BTW in `String.format` there is a variable of `%d` for decimal numbers like a score here.

src/main/java/com/epam/prejap/tetris/Tetris.java
rpiw: fancy way. I will remove it. I thought maybe some more things in Score class may be introduced in future, but well, I should ignore the future.

src/main/java/com/epam/prejap/tetris/Tetris.java
rpiw: Do we really need to test single print? Test would be a way longer than one-line print with local variable.

src/main/java/com/epam/prejap/tetris/Tetris.java
rpiw: Updated:
move printf to method where score is counted.

src/main/java/com/epam/prejap/tetris/Tetris.java
marcin-ogorzalek-EPAM: TBH I was after a name of the method here, not the method itself

src/main/java/com/epam/prejap/tetris/Tetris.java
marcin-ogorzalek-EPAM: Yes I would be that is why I left the decision to Tomasz

marcin-ogorzalek-EPAM:
please bump up version

src/main/java/com/epam/prejap/tetris/Tetris.java
marcin-ogorzalek-EPAM: There is also something like https://github.com/stefanbirkner/system-lambda to ease you in this task.

rpiw: test added

Issue #94, opened 2021-06-18
Add self-explanatory README.md and licenses to repo


rpiw:
Currently we do not have any license. License should be as open as licenses for all third-parties used in the repo. Any third-party license should be included in files if such one requires. README should explains why we do use those libs.

PR #96, opened 2021-06-18
94 add license

rpiw:
Add GNU GLP license. Update README. #94 
Edit: project depends (or will be in a few PR) on 3rd-party libs distributed under GNU GLP. Our license must not cause any conflicts with those.


TJB4epam (changes requested):
I don't see WHY we have added this license though? Wasn't it agreed to mention this and allow us to retrofit that decision later if we want to?

rpiw:
Added explanation in PR. Mention in readme that decision may be retrofitted in any time.

README.md
```diff
@@ -75,3 +75,8 @@ NOT HERE, NOT IN YOUR LOCAL REPOSITORY.
 1. Every successfully merged PR is worth 10 points.
 1. Every new, passing and meaningful test is worth 1 point.
 1. Every added and meaningful Javadoc is worth 1 point.
+
+
+## License
+Project is distributed under GNU Lesser General Public license.
+License may be changed in a future if all dependencies distributed under GNU LGP license were removed.
```
TJB4epam, line 82: Please add: Current GNU-licensed dependency is JNativeHook.

Also - bump the POM again. Then this can go in.

TJB4epam (changes requested):
Almost there!

PR #100, opened 2021-06-18
99 Add a comment on versions naming convention in pom.xml

AndriiVel:
This PR **adds a small comment** to the pom.xml regarding the naming convention for plugins and dependencies.

The versions already presented in pom.xml are **sorted** alphabetically (in accordance with this convention)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -16,9 +16,14 @@
         <test.jar.file.location>
             ${project.build.directory}${file.separator}${project.build.finalName}.jar
         </test.jar.file.location>
-        <testng.version>7.4.0</testng.version>
+        <!--
+            Please use this naming convention for plugin and dependency versions: <artifactId.version>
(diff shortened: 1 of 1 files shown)


MichalKulygin (commented):
TY @AndriiVel, very helpfull comment for people like me -> pre junior or prospective Epam employee 🤣 
kudos!

pom.xml
```diff
@@ -6,7 +6,7 @@
 
     <groupId>com.epam.prejap</groupId>
     <artifactId>tetris</artifactId>
-    <version>0.7</version>
+    <version>0.8</version>
```
TJB4epam, line 9: This is just a comment in the POM, not worthy of bumping the version. This saves you here as I just merged @cat-inGlasses changes. :)

TJB4epam (approved):
ok

Issue #93, opened 2021-06-18
BUG. Blocks are falling to the left side


AndriiVel:
The move randomizer works in a strange way. The blocks are falling mostly to the left side of the grid. Please, fix it.

Acceptance criteria:
* blocks fall to both sides equally
* there are tests proving this

AndriiVel:
Assuming that this issue is not yet taken, I would like to add my PR to fix it.

PR #106, opened 2021-06-21
93 Implement test for randomness with simple statistic checks.

rpiw:
Test if (according to #93) block are falling more often to left side. Current tests discredit the hypothesis. Implement check for standard deviations and check if sum of moves (where move left is negative move right) is close to zero.
@AndriiVel it was expected to work correctly, we use rng from standard library.


AndriiVel:
Wow! Sounds cool!👍
I'll check out your tests later, I'm curious 😁

AndriiVel:
I'm impressed by the tests, really!

However, I'm afraid some of them are biased (maybe not the best word, but cannot think of any better).

**TL;DR**
The bug is still there. The NONE and LEFT moves are selected indeed randomly.


From the JD the test method TestRandomPlayerRandomizer#testMovesSumUpCloseToZero seems to check the difference between number of times LEFT and number of times RIGHT moves appeared. The `counter` seems to accumulate only two values, presumably LEFT and RIGHT. That's the catch. Let's assume we add one more move (like, move block up). Then the test fails (approximate error message is "java.lang.AssertionError: Sum is -33452"). Well, maybe the tests can be rewritten to check the difference of occurrences of these three values...
But do we really have only three (after adding an extra move) moves?

Let's roll back the changes to initial state where we presumably had only LEFT and RIGHT moves.
A quick check to "enum Move" shows that we have not two, but three moves! NONE, LEFT, RIGHT.
Okay, then what does the "counter" stores then?
It stores first two moves. Move down and move left. We can prove it by adding a line `System.out.println("counter keys = " + new ArrayList<>(counter.keySet()));` to the TestRandomPlayerRandomizer#initializeFields() .
The output is "counter keys = [LEFT, NONE]"
If we go to the "enum Move" we can see, that the NONE and LEFT are first two constants. So the test method `testMovesSumUpCloseToZero` indeed prove that the selection is random... among two out of three values. The prove is cool and elaborated but the bug still sits there.

Okay, it looks like we have an "off by one" error here. Presumably somewhere in the so called (by me) randomizer. Our randomizer is called in class `RandomPlayer`. There is only one method and `random.nextInt()` is called with bound... 2! Because `Move.values().length - 1` will give you 2. The upper bound is excluded, so we get either 0 or 1; either NONE or LEFT.
If we add one more constant after RIGHT in "enum Move" then the RIGHT will be provided by `nextMove()`, however the newly added move will be still discarded (off by one error).


I don't know why you wanted to prove the randomness, assuming that the bug was spotted manually. Yes, It might have been just a coincidence. Although I ran the game 20 times before posting the Issue, just to be sure xD


Two ways to find this bug:
* manual tests by running the app and looking at it, let say, 20 times; 
* adding a logger and checking logs.

Both are not 100% bulletproof but in this case it worked.


Assuming that the issue didn't have any comments under it and that I spent some time on fixing the bug, I decided to open a PR to fix it.
I welcome your tests on randomness, they are really cool!

rpiw:
@AndriiVel {NONE=33268, RIGHT=33476, LEFT=33256}
Also, I fixed the test. Now it retrieves values for left and right moves directly by keys. 
And I see now that I did misunderstand you, as I thought you claimed problem was with randomness, but its not - the problem is what you described above: usage of nextInt with wrong arguments. I did not spot it, because I was checking tests with enum from my feature branch with 4 different types of moves which behaves as intended even with the bug. So simple proposal is to simply remove -1 from nextInt. Now I receive {NONE=24956, LEFT=24887, RIGHT=25151, TO_BOTTOM_NOW=25006}
So, excuse me inconvenience

rpiw:
Also, close, because #112 solve it in better way

AndriiVel:
I did the issue from the "black box" testing point of view. Without looking into the code, only from observing the game. I guess there will be times we will get descriptions with such detailed description =D

I hope you won't mind if I will use few ideas from your code in my PR. The `counter` is made in a really cool way!

rpiw:
sure, its gnu :P

Issue #110, opened 2021-06-21
Play Tetris music theme


AlexPasynkov:
### User story

`As a player,`
`I want to hear classic 8bit tetris music theme`
`so I can have nostalgia while I play`

### Acceptance criteria
1. When the game start I should hear song "Korobeiniki" in 8bit format

AlexPasynkov:
I take this

PR #111, opened 2021-06-21
Music player is added

AlexPasynkov:
Feature that was asked in #110 is now implemented and waiting for the request


src/main/java/com/epam/prejap/tetris/music/MusicPlayer.java
```diff
@@ -0,0 +1,34 @@
...
+            }
+        } else {
+            System.out.println("Error, please contact support for help");
+        }
+    }
+
+    public boolean fileExists(File file){
+        return file.exists();
+    }
+}
```
marcin-ogorzalek-EPAM, line 34: No new line at end of file :(

src/main/java/com/epam/prejap/tetris/music/MusicPlayer.java
```diff
@@ -0,0 +1,34 @@
...
+     * This method is void and does not take any parameters
+     * It is invoked in Tetris class, during game start of the game
+     */
+    public void playMusic(){
+        file = new File("src/main/resources/tetris8bit.wav");
+        AudioInputStream audioStream;
+        if(fileExists(file)){
+            try {
+                audioStream = AudioSystem.getAudioInputStream(file);
+                Clip clip = AudioSystem.getClip();
```
marcin-ogorzalek-EPAM, line 20: Shouldn't we close somehow this resource at the end (regardless of exception or not)? Moreover, isn't this one implements `AutoCloseable` interface?

src/main/java/com/epam/prejap/tetris/music/MusicPlayer.java
```diff
@@ -0,0 +1,34 @@
...
+                clip.open(audioStream);
+                clip.start();
+            } catch (UnsupportedAudioFileException | LineUnavailableException | IOException e) {
+                e.printStackTrace();
```
marcin-ogorzalek-EPAM, line 24: Is this a good practice?

src/main/java/com/epam/prejap/tetris/music/MusicPlayer.java
```diff
@@ -0,0 +1,34 @@
...
+            }
+        } else {
+            System.out.println("Error, please contact support for help");
```
marcin-ogorzalek-EPAM, line 27: It seems kind of like an error message, yet it goes in a normal output stream. I would suggest at least `System.err.println`

src/main/java/com/epam/prejap/tetris/music/MusicPlayer.java
```diff
@@ -0,0 +1,34 @@
...
+public class MusicPlayer {
+
+    File file;
+    /**
+     * This method is void and does not take any parameters
+     * It is invoked in Tetris class, during game start of the game
+     */
+    public void playMusic(){
+        file = new File("src/main/resources/tetris8bit.wav");
+        AudioInputStream audioStream;
```
marcin-ogorzalek-EPAM, line 16: This one also implements `AutoCloseable`.

src/test/java/com/epam/prejap/tetris/music/musicTest.java
```diff
@@ -0,0 +1,17 @@
...
+
+
+import org.testng.annotations.Test;
+
+import java.io.File;
+
+import static org.testng.Assert.assertTrue;
+
+@Test
+public class musicTest {
```
marcin-ogorzalek-EPAM, line 11: Java naming convention.

src/test/java/com/epam/prejap/tetris/music/musicTest.java
```diff
@@ -0,0 +1,17 @@
...
+    public static final String MUSIC_PATH = "src/main/resources/tetris8bit.wav";
```
marcin-ogorzalek-EPAM, line 12: Well for tests, there rather should be another file in `src/test/resources/`.

src/test/java/com/epam/prejap/tetris/music/musicTest.java
```diff
@@ -0,0 +1,17 @@
...
+    public void musicFileShouldBeStoredInTheResourcesFolder(){
+        File file = new File(MUSIC_PATH);
+        assertTrue(file.exists(),"Music file is missing");
```
marcin-ogorzalek-EPAM, line 15: This tests rather a method `file.exists()` no a presence of the file itself. What if the method would be buggy and regardless of that the file would be present in the directory.

src/test/java/com/epam/prejap/tetris/music/musicTest.java
marcin-ogorzalek-EPAM: Interconnected with https://github.com/EPAM-Online-Courses/tetris/pull/111/files#r655989553

src/main/java/com/epam/prejap/tetris/Tetris.java
```diff
@@ -13,15 +14,18 @@
     private final Playfield playfield;
     private final Waiter waiter;
     private final Player player;
+    private final MusicPlayer musicPlayer;
 
-    public Tetris(Playfield playfield, Waiter waiter, Player player) {
+    public Tetris(Playfield playfield, Waiter waiter, Player player, MusicPlayer musicPlayer) {
```
marcin-ogorzalek-EPAM, line 19: I think that with the number of the parameters we are approaching the threshold where some creational design pattern could be introduced (rather in a new issue).

src/main/java/com/epam/prejap/tetris/music/MusicPlayer.java
AlexPasynkov: I think not. At the start I had 3 different catch blocks. But Idea gave me 3 warnings that I should combine it in one catch block. I thought that IDEA knows better

src/main/java/com/epam/prejap/tetris/music/MusicPlayer.java
```diff
@@ -0,0 +1,34 @@
+package com.epam.prejap.tetris.music;
+
+import javax.sound.sampled.*;
+import java.io.File;
+import java.io.IOException;
+
+public class MusicPlayer {
+
+    File file;
```
marcin-ogorzalek-EPAM, line 9: why not `private`?

src/main/java/com/epam/prejap/tetris/music/MusicPlayer.java
```diff
@@ -0,0 +1,34 @@
...
+    /**
+     * This method is void and does not take any parameters
+     * It is invoked in Tetris class, during game start of the game
+     */
+    public void playMusic(){
+        file = new File("src/main/resources/tetris8bit.wav");
```
marcin-ogorzalek-EPAM, line 15: Maybe there should be a constructor that takes the path as a parameter.

marcin-ogorzalek-EPAM (changes requested):
Java naming convention hurts our hearts the most.

src/main/java/com/epam/prejap/tetris/music/MusicPlayer.java
AlexPasynkov: Yes, it should be error. Already changed

src/main/java/com/epam/prejap/tetris/music/MusicPlayer.java
marcin-ogorzalek-EPAM: I'm not talking about multi-catch. This is fine. I'm talking about `printStackTrace()`. There was something about it during the proper lecture.

src/test/java/com/epam/prejap/tetris/music/musicTest.java
AlexPasynkov: Ohh, I really called a class with a small letter??? Oh my, fixed.

src/test/java/com/epam/prejap/tetris/music/musicTest.java
AlexPasynkov: Can I test if the method play music was invoked when the game started?

src/main/java/com/epam/prejap/tetris/Tetris.java
TJB4epam: Definitely. Since Tetris with music may be a thorny issue, I'd like a control for muting and I'd like to have two Tetris classes, one with music and one without.

After all - audio uses some CPU - for high-stakes games, we may want that CPU for something else... ;-) ;-)

Seriously though: muting is a must. Having a music player as an OPTION is a must as well. Whether you do it via setter or a design pattern is something for you to choose. :-)

src/test/java/com/epam/prejap/tetris/music/musicTest.java
TJB4epam: Yes. And there are tests for audio. You can test that proper JDK / library method was invoked, and assume that the method does the job (talks to OS, passes audio stream, OS does its job).

TJB4epam (changes requested):
POM version?

Audio files are files, true. But when called from a JAR, they're no longer files, they're resources. 

More tests required.

TJB4epam:
And - one more important thing - I LOVE the idea. :-)

PR #103, opened 2021-06-19
7 rotation of blocks

MichalKulygin:
Added clock-wise block rotation feature.

PR #62, opened 2021-06-05
14 send block to bottom immediatly

rpiw:
Implement keyboard handling with JNativeHook library. Allow user to control the block with 'h', 'l' and arrow down buttons. To run as a human player send **human** string as a parameter to jar. Introduce simple logging facility with log4j library for debugging. 

To sum up: solve #14 issue

EDIT by TJB, for clarity put here:

1. I want to use `j` for moving to the bottom - more Vim-like, and more in tune with `h` and `l`. 
2. I also propose to introduce logging as a separate feature - we have feature creep here. I wouldn't want to have logging just like so as it's actually fairly important. Thus here I'd like to see any logging, just introduced, but am equally fine with no logging here at all, leaving true logging for later as a separate issue. See #79

PR #112, opened 2021-06-21
93 fix bug blocks are falling to the left side

AndriiVel:
This PR fixes the bug in issue #93  where the blocks were falling mostly to the left side of the grid.

The bug was in moves selection, when we were getting only moves down and left for blocks. The right move was never accessed.

The fix also specifies the moves available for a RandomPlayer. This will prevent for possible bugs in the future, when we will add more moves.


src/main/java/com/epam/prejap/tetris/game/Move.java
```diff
@@ -21,4 +21,13 @@ public static Move of(int key) {
         }
         return NONE;
     }
+
+    /**
+     * Provides moves used by {@link com.epam.prejap.tetris.player.RandomPlayer}
+     *
+     * @return an array of possible moves
+     */
+    public static Move[] getRandomPlayerMoves() {
```
marcin-ogorzalek-EPAM, line 30: Isn't this method actual doing the same thing as `values()` building to enum type? What is the reasoning behind coding this explicitly?

src/test/java/com/epam/prejap/tetris/player/RandomPlayerTest.java
```diff
@@ -0,0 +1,41 @@
...
+
+@Test(groups = "players")
+public class RandomPlayerTest {
+
+    private static final int REPETITIONS = 1000;
+
+    public void nextMoveReturnsAllAvailableMoves() {
+        RandomPlayer player = new RandomPlayer();
+        Set<Move> movesProduced = new HashSet<>();
+        Set<Move> actualMoves = new HashSet<>(Arrays.asList(Move.getRandomPlayerMoves()));
```
marcin-ogorzalek-EPAM, line 20: This can be simplified with `Set.of`. There is a version for varargs and varargs are under the hood arrays. Thus if I'm correct an array can be directly passed to this method.

marcin-ogorzalek-EPAM: BTW is those are actual moves? Made by who? Maybe those are all possible moves.

src/test/java/com/epam/prejap/tetris/player/RandomPlayerTest.java
```diff
@@ -0,0 +1,41 @@
...
+
+import static org.testng.Assert.assertEquals;
+
+@Test(groups = "players")
+public class RandomPlayerTest {
+
+    private static final int REPETITIONS = 1000;
+
+    public void nextMoveReturnsAllAvailableMoves() {
+        RandomPlayer player = new RandomPlayer();
```
marcin-ogorzalek-EPAM, line 18: As we are using an instance of the RandomPlayer in both tests, its initialisation could be moved to some setup method annotated with `@BeforeMethod`

src/main/java/com/epam/prejap/tetris/game/Move.java
AndriiVel: It is doing the same, as for now. However in case we add more moves then the RandomPlayer would also choose them. 
Possible moves that may appear:  "rotate the block clockwise" in PR #103 and "send block immediately to the bottom" in PR #62 .
I don't think this is what RandomPlayer should do.

Of course this is a game design decision, please tell me if you want such moves to be selected by Random player. I would just roll back the changes to previous commit in this PR.

src/test/java/com/epam/prejap/tetris/player/RandomPlayerTest.java
AndriiVel: Will change, thank you.
As for the naming, maybe "possible moves" is a better name. Will change. TY!

src/test/java/com/epam/prejap/tetris/player/RandomPlayerTest.java
AndriiVel: Will do 👍

src/test/java/com/epam/prejap/tetris/player/RandomPlayerTest.java
```diff
@@ -0,0 +1,41 @@
...
+        Set<Move> movesProduced = new HashSet<>();
+        Set<Move> actualMoves = new HashSet<>(Arrays.asList(Move.getRandomPlayerMoves()));
+
+        for (int i = 0; i < REPETITIONS; i++) {
+            movesProduced.add(player.nextMove()
+                    .orElse(Move.NONE));
+        }
```
marcin-ogorzalek-EPAM, line 25: As this code of block is the same in both tests maybe it should be enclosed in `private` method used there.

marcin-ogorzalek-EPAM (changes requested):
There are some improvements to be made.

src/main/java/com/epam/prejap/tetris/game/Move.java
marcin-ogorzalek-EPAM: Ok, you convinced me. It may stay (if Tomasz does not mind).

src/test/java/com/epam/prejap/tetris/player/RandomPlayerTest.java
AndriiVel: Good idea! Will do

AndriiVel:
Project version was changed to 0.11 according to MS Teams post by Tomasz in Lectures links.

src/test/java/com/epam/prejap/tetris/player/RandomPlayerTest.java
```diff
@@ -0,0 +1,76 @@
...
+
+        assertEquals(movesProduced, possibleMoves);
+    }
+
+    public void numberOfUniqueMovesReturnedByNextMoveShouldBeEqualToNumberOfAvailableMoves() {
+        Set<Move> movesProduced = new HashSet<>();
+
+        makeMoves(movesProduced);
+
+        assertEquals(movesProduced.size(), Move.getRandomPlayerMoves().length);
```
TJB4epam, line 38: Moves will change in a moment, heads-up!

src/test/java/com/epam/prejap/tetris/player/RandomPlayerTest.java
```diff
@@ -0,0 +1,76 @@
...
+import org.testng.annotations.Test;
+import org.testng.asserts.SoftAssert;
+
+import java.util.HashSet;
+import java.util.Set;
+
+import static org.testng.Assert.assertEquals;
+
+@Test(groups = "players")
+public class RandomPlayerTest {
```
TJB4epam, line 14: That's integration tests, cause you're testing random and have plenty of repetitions. BTW, TestNG has `invocationCount` that may be of use here.

TJB4epam (changes requested):
Update with newest merges, take a look at invo count in TNG.

src/test/java/com/epam/prejap/tetris/player/RandomPlayerTest.java
AndriiVel: And they might change in the future again. That's why I propose to use for RandomPlayer only three basic moves: left, right, down.

src/test/java/com/epam/prejap/tetris/player/RandomPlayerTest.java
AndriiVel: Ok. Will mark it appropriately.

AndriiVel: Three tests are now using the `invocationCount` and `successPercentage`. 

Kudos to @asilenko for helping with figuring out how to use them!

src/main/java/com/epam/prejap/tetris/game/Move.java
pkierat-epam: Dunno if Tomasz has some objections, but I do... AFAICS, we already have more than one implementation of the `Player` interface, and HumanPlayer actually shouldn't care about the random moves. In fact, random moves are only important to `RandomPlayer`. Therefore, putting this logic in a common class is a violation of SRP. The place for it is the `RandomPlayer` class.

AndriiVel: Thank you, @pkierat-epam !
Changed!

src/main/java/com/epam/prejap/tetris/player/RandomPlayer.java
```diff
@@ -1,16 +1,26 @@
...
+     * @return a randomly selected move from {@link #getAllPossibleMoves()}
+     */
     @Override
     public Optional<Move> nextMove() {
-        return Optional.of(Move.values()[random.nextInt(Move.values().length - 1)]);
+        return Optional.of(getAllPossibleMoves()[random.nextInt(getAllPossibleMoves().length)]);
+    }
+
+    private Move[] getAllPossibleMoves() {
+        return new Move[]{Move.NONE, Move.LEFT, Move.RIGHT};
```
pkierat-epam, line 23: NON-BLOCKER: Having that as a method which always returns a new array containing exactly the same elements is a waste of memory and CPU cycles. A better option is to use EnumSet stored in a field.

AndriiVel: Done

PR #114, opened 2021-06-22
113 Follow-up to S-block

asilenko:
Follow-up to S-block #113
Solved. Block in BlockFeed and in Test are now in alphabetical order. 
In addition:
- remove duplicated IBlock from BlockFeed,
- refactor test class name to plural.

PR #115, opened 2021-06-22
new chages with L  block

DzmitryTatarski:


cat-inGlasses:
I think, there is no need in LBlockTest, as all what is tested there will be tested as you added entry BlockShapeData.

src/test/java/com/epam/prejap/tetris/block/BlockShapeData.java
```diff
@@ -46,6 +46,11 @@
                 {1, 1, 1},
                 {1, 0, 1}
         });
+        put(LBlock::new, new byte[][]{
+                {1, 0},
+                {1, 0},
+                {1, 1},
+        });
```
cat-inGlasses, line 53: I think, here, also be better to add block's test data in alphabetical order, as where done in this PR - [113 Follow-up to S-block #114](https://github.com/EPAM-Online-Courses/tetris/pull/114)

TJB4epam (changes requested):
Changes to be done:

1. PRs should be described, so reviewers know what the change is about, what you want them to focus on, what - roughly - you did here. On Wednesday we are likely to close all PRs that don't have descriptions. Just so you know in advance.
2. TestNG on compile scope is strange. Why?
3. POM has no version update. See Teams for which version that should be.
4. No need for version in the JD. We do not use it in Tetris. If you wish to use it, please, use the version as per POM. Version 1.0 is something we are yet to reach.
5. The way to test the block is now different - see changes done by @asilenko  and @cat-inGlasses 

There are more, but for starters that should do.

marcin-ogorzalek-EPAM:
No PR description, conflicts, the version to bump up (at least .16)

